PR #5, opened 2024-10-29
feat: Beginning to flesh out the interface

lewisjared:
## Description

Flesh out a basic interface for a metric.

The goal is to provide an interface that is:
* flexible to support a range of different ways of calling the benchmarking package
* provides the ability to include some glue code to do REF-specific "stuff"
* include type-hints to be explicit about what is available and what output is expected
* easily testable
* be extensible in future

## Checklist

- [x] Tests added


packages/ref-metrics-example/src/ref_metrics_example/example.py
```diff
@@ -0,0 +1,115 @@
...
+    """
+    input_files = dataset.glob("*.nc")
+
+    xr_ds = xr.open_mfdataset(list(input_files), combine="by_coords", chunks=None, use_cftime=True)
+
+    annual_mean = xr_ds.resample(time="YS").mean()
+    return annual_mean.mean(dim=["lat", "lon"], keep_attrs=True)
+
+
+def format_cmec_output_bundle(dataset: xr.Dataset) -> dict[str, Any]:
```
znichollscr, line 36: (Maybe I'm jumping ahead and you're planning on doing this already but not yet while you're prototyping, if so just ignore me). Is it worth creating a class for the output rather than using the ambiguous `dict[str, Any]`? We can always serialise to a dict if we need, but having a class could help communicate what is actually needed.

znichollscr: Just saw this: https://github.com/CMIP-REF/cmip-ref/pull/5/files#diff-2f33f70c0be4a6d039d01398ce350b06f838a71a347972a772299bfab43842e7R58

Ignore me, nice stuff

znichollscr: I would have resolved the thread already, but for some reason I cannot 🙃

lewisjared:
I'm going to merge this now so this doesn't block anything. We can update fields etc easily at this stage

Issue #10, opened 2024-11-07
Input and obs requirements


lewisjared:
## The problem
As a metric provider, I want to be able to define the datasets that my metrics require in a consistent manner, so that the metric is triggered when any of these datasets are added/modified/deleted/retracted.

These requirements will form the basis solving which metrics need to be calculated on an incoming dataset.

Some metrics will require operating on a single dataset while others will require multiple e.g. a multi-model mean.
We have to have some mechanism for handling both cases.
Observation data should follow a similar convention.

Filter by facets?

How freeform should this be? Do we need to be flexible to the availability of different facet i.e from CMIP7-FT and input4MIPs

## Definition of "done"
* [ ] Create some types to encapsulate the required info
* [ ] Add these to the metric protocol

lewisjared:
@nocollier @lee1043 @bouweandela Am I being naive if I assume that the requirements for a given metric are reduced down to a set of filter operations using facets? It seems like a somewhat normal pattern although both ESMValTool and CMEC tend to use a pattern of explicit inclusion i.e. models are explicitly specified rather than globbing. 

Since we can't really have a person in the loop, I think we need to rely on globbing and some early QA/QC by ESGF (might not happen in FT), then either failing to calculate or capture an erroneous result which might be obvious when compared to other input datasets, which is indicative that that input dataset isn't like the others. Failing is probably the better approach if something is obviously wrong. I think the decision to fail lies with the metric service.

I think we might also need a groupby field to determine how jobs are triggered. 

Assuming that we had a big pandas dataframe containing metadata about each source dataset (pretty much an intake data catalog). This df would include the ESGF facets + some extra info not available via the ESGF API (branch times etc). For a given metric, the filters that is provides will be used to reduce this dataframe down to just the datasets that are applicable to it. Then this dataframe could be grouped the unique values in a set of columns. Each group would result in an execution of the metric.

Psuedo code might be more helpful compared to a wall of words...
```python
def run_metric(df: pd.DataFrame):
    print(f"running with: {df}")

my_datasets: pd.DataFrame = get_available_datasets()

filters = [{
	    "variable_id": ["tas"]
		"experiment_id": ["historical"]
	},
	{
	    "keep": False,
		"model": "XYZ"
	}]

for filter in filters:
 	my_datasets = apply_filters(my_datasets, filter)

my_datasets.groupby(columns=["model", "variant_id"]).apply(run_metric)
```

[...]

PR #15, opened 2024-11-20
Update the metric protocol to include the concept of inputs

lewisjared:
## Checklist

- [x] Tests added


lewisjared:
This is a potential config interface for #10

packages/ref-core/src/ref_core/metrics.py
```diff
@@ -108,18 +165,14 @@ class Metric(Protocol):
     but multiple providers can implement the same metric.
     """
 
-    # input_variable: list[VariableDefinition]
+    inputs: list[DataRequirement]
```
lewisjared, line 168: Should this be a list of requirements or split them out?
```suggestion
    source_requirements: DataRequirement | None
    reference_requirements: DataRequirement | None
    metric_requirements: DataRequirement | None
```

As a non-expert the data sources are interchangeable as a change in any dependent data would trigger a calculation, but as a metric developer there might be other opinions. IMHO how the trigger information is presented to the metric calculation is somewhat independent to this.

Pros for single field: No schema changes if additional dimensions of datasets are added, supports the concept of multiple input sources (CMIP6 + CMIP7).
Cons: requires knowing to add additional dependencies on different data types rather than it being more explicit

Pros for multiple: More explicit, don't need to deal with the potential of multiple duplicate sources.
Cons: Less flexible

lewisjared:
@mikapfl Still a WIP but it is at a good point to get some feedback. #10 for context

packages/ref-core/src/ref_core/constraints.py
```diff
@@ -0,0 +1,148 @@
...
+
+from ref_core.exceptions import ConstraintNotSatisfied
+
+
+@runtime_checkable
+class GroupValidator(Protocol):
+    """
+    A constraint that must be satisfied when executing a given metric run.
+
+    All constraints must be satisfied for a given group to be run.
```
mikapfl, line 15: > group

I think it would be good to define the exact meaning of a group somewhere. I didn't find anything in the docs so far, and it seems the details are crucial here.

packages/ref-core/src/ref_core/constraints.py
```diff
@@ -0,0 +1,148 @@
...
+        :
+            Whether the constraint is satisfied
+        """
+        ...
+
+
+@runtime_checkable
+class GroupOperation(Protocol):
+    """
+    An operation to perform on a group of datasets resulting in a new group of datasets.
```
mikapfl, line 41: > group of datasets

I assume the group of datasets here is not the same as the group above? Otherwise, I'm probably struggling with the language "the group will not be executed" because datasets can't be executed, right?

packages/ref-core/src/ref_core/constraints.py
```diff
@@ -0,0 +1,148 @@
...
+        """
+        Perform an operation on the group of datasets.
+
+        If this operation modifies the group, the modified group should be returned.
+        If this operation fails, a ConstraintNotSatisfied exception should be raised.
+
+        Parameters
+        ----------
+        group
+            A group of datasets that is being validated.
```
mikapfl, line 54: Can you specify somewhere (maybe just in words here or somewhere to not start down the path of pandas typing) what info the group df needs to contain, i.e. what columns are available, I assume it is long format?

packages/ref-core/src/ref_core/constraints.py
```diff
@@ -0,0 +1,148 @@
...
+Constraint = GroupOperation | GroupValidator
+"""
+A constraint that must be satisfied when executing a given metric run.
+
+This can include operations that are applied to a group of datasets which may modify the group,
+but may also include validators that check if the group satisfies a certain condition.
+
+All constraints must be satisfied for a given group to be run.
+If any constraint is not satisfied, the group will not be executed.
+"""
```
mikapfl, line 80: Do I understand correctly?

A Constraint operates on a Group of datasets (which contains metadata about datasets only, no data) and returns a constrained Group of datasets, which is a subset of the input Group of datasets. Alternatively, it can signal that the input Group is not valid at all.

As a special case, a GroupValidator will not constrain the input Group of datasets, instead it can only signal that the input Group is not valid. Every GroupValidator could be implemented as a GroupOperation instead, but a GroupValidator might be easier to understand.

packages/ref-core/src/ref_core/constraints.py
```diff
@@ -0,0 +1,148 @@
...
+            raise ConstraintNotSatisfied(f"Constraint {constraint} not satisfied for {dataframe}")
+    except ConstraintNotSatisfied:
+        logger.debug(f"Constraint {constraint} not satisfied for {dataframe}")
+        return None
+
+    return updated_group
+
+
+@frozen
+class RequiredFacets:
```
mikapfl, line 118:
```suggestion
class RequireFacets:
```

For `SelectParentExperiment` you use imperative present, I think that would be good to use everywhere in the Constraint names.

packages/ref-core/src/ref_core/constraints.py
```diff
@@ -0,0 +1,148 @@
...
+            logger.warning(f"Dimension {self.dimension} not present in group {group}")
+            return False
+        return all(value in group[self.dimension].values for value in self.required_facets)
+
+
+@frozen
+class SelectParentExperiment:
+    """
+    Include a dataset's parent experiment in the selection
+    """
```
mikapfl, line 140: Oh, so the Group of datasets returned from a GroupOperation does not need to be a subset of the input Group of datasets? Fancy. So it could also contain datasets from the data_catalog?

mikapfl: This raises the question what happens with multiple Constraints. Do they need to be ordered, so that the GroupOperations are applied in order?

packages/ref-core/src/ref_core/datasets.py
```diff
@@ -8,3 +10,25 @@ class SourceDatasetType(enum.Enum):
...
+    The keys are the metadata fields to filter on, and the values are the values to filter on.
+    If multiple values are provided for a single field, the filter will be applied as an OR operation.
+    Multiple filters are applied as an AND operation.
+    """
+    keep: bool = True
+    """
+    Whether to keep or remove datasets that match the filter.
+
+    If true (default), datasets that match the filter will be kept else they will be removed.
+    """
```
mikapfl, line 34: Maybe it is worth to have a `IncludeFacetFilter` and another `ExcludeFacetFilter`. I sometimes tend to overlook the innocent `keep=False` after a long filter spec e.g. in scmdata.

packages/ref-core/src/ref_core/metrics.py
```diff
@@ -87,6 +91,79 @@ class TriggerInfo:
...
+    Definition of the input datasets that a metric requires to run.
+
+    This is used to determine the different metric executions that are required to be performed
+    with the current data catalog.
+
+    The data catalog is filtered according to the `filters` field,
+    then grouped according to the `group_by` field,
+    and then each group is checked that it satisfies the `constraints`.
+    Each such group will be processed as a separate execution of the metric.
+    """
```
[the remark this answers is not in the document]
mikapfl: Maybe we could also just let people provide a grouping function? Most may be happy using pandas' groupby, but some might want to do fancy stuff, and why not?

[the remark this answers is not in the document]
mikapfl: Also, not to derail this, but depending on the people we expect to write DataRequirements, SQL could also be a way to specify filters and groupbys. SQL has a rich language for this stuff, we are probably selecting from a database anyway, and sqlite is faster than anything we can implement using pandas. Much faster, if we include the correct indexes.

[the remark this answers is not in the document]
mikapfl: But GroupOperations might not be easily possible in SQL, e.g. getting the parent.

mikapfl (changes requested):
I've got a lot of questions, probably because I haven't worked a lot with this kinds of datasets, yet. But maybe a useful perspective as a somewhat-outsider.

packages/ref-core/src/ref_core/metrics.py
```diff
@@ -87,6 +91,79 @@ class TriggerInfo:
...
+
+    source_type: SourceDatasetType
```
bouweandela, line 108: Isn't this just another facet? Or is it treated separately because you plan to validate the facet names themselves at some point?

lewisjared: I think the answer is maybe. I could imagine non-CMIP data being filtered in a similar way e.g. previous executions of the metrics. I'm not certain if we want to lump these different data catalogs together or keep them separate. They will have quite different facets.

If we opt to split the requirements by type (https://github.com/CMIP-REF/cmip-ref/pull/15#discussion_r1849585098) then this field is superfluous.

packages/ref-core/src/ref_core/constraints.py
lewisjared: > So it could also contain datasets from the data_catalog?

Yes. We need a lot of flexibility here

> Do they need to be ordered, so that the GroupOperations are applied in order?
They are ordered and are applied in that same order. If any fail then the group is skipped.

I think I documented that, but I'll double check

packages/ref-core/src/ref_core/constraints.py
lewisjared: Yup sure. It seems like there needs to be some vocabulary definitions in the docs

FYI: A data catalog has a row per file, and the columns are defined by source type: https://github.com/CMIP-REF/cmip-ref/blob/9ea6e99ace037c2ef1a6ae26ce43d52915cd2961/packages/ref/src/ref/datasets/cmip6.py

packages/ref-core/src/ref_core/constraints.py
lewisjared: I would write it as:

A Constraint operates on a Group of datasets (which contains metadata about datasets only, no data). There are two types of functionality that a Constraint may implement:
* A GroupOperation modifies a Group of datasets, potentially including additional datasets from the complete data catalog
* A GroupValidator can be used to check if the Group of datasets should be executed.

A Constraint can implement one or both of those interfaces in which case the GroupOperation is performed before the GroupValidator. Any Constraints that fail validation or raise exceptions will cause the Group of datasets to not be marked for execution.

> Every GroupValidator could be implemented as a GroupOperation instead, but a GroupValidator might be easier to understand.

Correct. If the GroupOperation raises a `ConstraintNotSatisfied` it would effectively be the same as a GroupValidator. Clearly delineating mutating the Group of datasets from validation feels nicer to me, but it might be adding unnecessary complexity.

packages/ref-core/src/ref_core/metrics.py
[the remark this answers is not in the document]
lewisjared: > SQL could also be a way to specify filters and groupbys

There might be optimisations that are used in the backend to load the appropriate  data catalog/groups. Currently this looks like [this](https://github.com/CMIP-REF/cmip-ref/blob/17d5119b2ea96ff61328e2cd52e2da587476f9a5/packages/ref/src/ref/datasets/cmip6.py#L220) with no filters. I've already got a TODO in there suggesting that this might not be the best approach.

The author of these DataRequirements will be the metric package authors (a mix of Scientist and RSEs). My gut says that a micro DSL will be easier to understand and have more guardrails than freeform SQL. Some effort will have to go into sane error messages (column x doesn't exist), logging and I'd probably suggest a CLI entrypoint to perform the solving in a stepwise manner to help verify what is happening. 

This will also require a how-to-guide (notebook) for the docs

It doesn't exclude having a SQLDataRequirement for expert 😉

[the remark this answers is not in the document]
lewisjared: > SQL could also be a way to specify filters and groupbys

There might be optimisations that are used in the backend to load the appropriate data catalog/groups. Currently this looks like [this](https://github.com/CMIP-REF/cmip-ref/blob/17d5119b2ea96ff61328e2cd52e2da587476f9a5/packages/ref/src/ref/datasets/cmip6.py#L220) with no filters. I've already got a TODO in there suggesting that this might not be the best approach.

packages/ref-core/src/ref_core/constraints.py
lewisjared: https://github.com/CMIP-REF/cmip-ref/blob/requirements/docs/how-to-guides/running-metrics-locally.py

packages/ref-core/src/ref_core/datasets.py
lewisjared: https://github.com/CMIP-REF/cmip-ref/issues/28

lewisjared:
@bouweandela This is already getting a bit bit. Happy to tweak this to better match your breakdown in #15, but it might occur in a follow up MR

docs/how-to-guides/dataset-selection.py
```diff
@@ -0,0 +1,216 @@
...
+#     text_representation:
+#       extension: .py
+#       format_name: percent
+#       format_version: '1.3'
+#       jupytext_version: 1.16.4
+#   kernelspec:
+#     display_name: Python 3 (ipykernel)
+#     language: python
+#     name: python3
+# ---
```
mikapfl, line 13: I couldn't run this notebook because my `data_catalog` is empty. Is there some set-up that I should do before I can run it?

packages/ref-core/src/ref_core/constraints.py
```diff
@@ -39,13 +39,19 @@ def validate(self, group: pd.DataFrame) -> bool:
...
+        Mutating the input group may result in unexpected behaviour.
     """
 
     def apply(self, group: pd.DataFrame, data_catalog: pd.DataFrame) -> pd.DataFrame:
         """
         Perform an operation on the group of datasets.
 
-        If this operation modifies the group, the modified group should be returned.
+        A new group of datasets should be returned if modifications are required,
+        and the input group should not be modified.
```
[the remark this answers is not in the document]
lewisjared: That is the current behaviour.

docs/how-to-guides/dataset-selection.py
lewisjared: You should pull the data needed for the tests. This will setup esgpull in the current directory

```
uv run esgpull self install $PWD/.esgpull
make fetch-test-data
```

There isn't a super happy path to getting the local config setup for the REF yet, but what I tend to do is:

```
uv run ref config list > $PWD/.ref/ref.toml
export REF_CONFIGURATION=$PWD/.ref
```
then edit `$PWD/.ref/ref.toml` to fix the paths and database url (they will default to somewhere in your home directory) and set `paths.allow_out_of_tree_datasets` to true. This does three things, creates an in-source config file,  sets the `ref` cli to use the new config and allows for datasets to be ingested from outside of `paths.data`.

My ref.toml file for reference:
```
[paths]
data = "/Users/jared/code/CMIP-REF/cmip-ref/.ref/data"
log = "/Users/jared/code/CMIP-REF/cmip-ref/.ref/log"
tmp = "/Users/jared/code/CMIP-REF/cmip-ref/.ref/tmp"
allow_out_of_tree_datasets = true

[db]
database_url = "sqlite:////Users/jared/code/CMIP-REF/cmip-ref/.ref/db/ref.db"
run_migrations = true
```

Then ingest the downloaded datasets into the database
```
uv run ref ingest --source-type cmip6 $PWD/.esgpull/data
```

On the first run this will create a sqlite file according to `db.database_url` and apply any migrations, before walking the directory provided to find that available datasets which are then added to the db.

mikapfl: Nice, thanks, that worked. Is it worth including this somewhere in the developer docs?

mikapfl (approved):

docs/how-to-guides/dataset-selection.py
lewisjared: Yup definitely. I've made `allow_out_of_tree_datasets` default to true for now to make it easier setup to the test suite. This should probably be false for users but we are a while away from worrying about the happy path for users

lewisjared:
Thanks for the feedback @mikapfl and @bouweandela. I'm going to merge this now. My goal is to come up with an end-to-end workflow asap and then we can revisit some of these assumptions when we implement some metrics.

Issue #32, opened 2024-12-05
Implement and test ESMValTool integration


lewisjared:
# Descripition

Build out the ESMValTool integration within the REF.

# Definitions of done

* [x] Add a `ref-metrics-esmvaltool` package
* [ ] Setup a conda file with ESMValTool dependencies
* [ ] Metric a
* [ ] Feedback on interfaces
* [ ] Metric b
* [ ]

bouweandela:
Related to https://github.com/ESMValGroup/ESMValTool/issues/3824

lewisjared:
Marking the board item as done as we have an example implementation

PR #40, opened 2024-12-12
Add ESMValTool example metric

bouweandela:
## Description

Add an example ESMValTool metric. First step towards #32.

## Checklist

- [ ] Tests added

### DIFF
--- a/packages/ref-metrics-esmvaltool/src/ref_metrics_esmvaltool/example.py
+++ b/packages/ref-metrics-esmvaltool/src/ref_metrics_esmvaltool/example.py
@@ -1,0 +1,108 @@
+from typing import Any
+
+import xarray as xr
+from ref_core.datasets import FacetFilter, SourceDatasetType
+from ref_core.metrics import DataRequirement, Metric, MetricExecutionDefinition, MetricResult
+from ruamel.yaml import YAML
+
+from ref_metrics_esmvaltool.recipe import dataframe_to_recipe, load_recipe, run_recipe
+
+yaml = YAML()
+
+
+def format_cmec_output_bundle(dataset: xr.Dataset) -> dict[str, Any]:
+    """
+    Create a simple CMEC output bundle for the dataset.
+
+    Parameters
+    ----------
+    dataset
+        Processed dataset
+
+    Returns
+    -------
+        A CMEC output bundle ready to be written to disk
+    """
+    # TODO: Check how timeseries data are generally serialised
+    cmec_output = {
+        "DIMENSIONS": {
+            "dimensions": {
+                "source_id": {dataset.attrs["source_id"]: {}},
+                "region": {"global": {}},
+                "variable": {"tas": {}},
+            },
+            "json_structure": [
+                "model",
+                "region",
+                "statistic",
+            ],
+        },
+        # Is the schema tracked?
+        "SCHEMA": {
+            "name": "CMEC-REF",
+            "package": "example",
+            "version": "v1",
+        },
+        "RESULTS": {
+            dataset.attrs["source_id"]: {"global": {"tas": 0}},
+        },
+    }
+
+    return cmec_output
+
+
+class GlobalMeanTimeseries(Metric):
+    """
+    Calculate the annual mean global mean timeseries for a dataset
+    """
+
+    name = "Global Mean Timeseries"
+    slug = "global-mean-timeseries"
+
+    data_requirements = (
+        DataRequirement(
+            source_type=SourceDatasetType.CMIP6,
+            filters=(FacetFilter(facets={"variable_id": ("tas",)}),),
+            # Add cell areas to the groups
+            # constraints=(AddCellAreas(),),
+            # Run the metric on each unique combination of model, variable, experiment, and variant
+            group_by=("source_id", "variable_id", "experiment_id", "variant_label"),
+        ),
+    )
+
+    def run(self, definition: MetricExecutionDefinition) -> MetricResult:
+        """
+        Run a metric
+
+        Parameters
+        ----------
+        definition
+            A description of the information needed for this execution of the metric
+
+        Returns
+        -------
+        :
+            The result of running the metric.
+        """
+        # Load recipe and clear unwanted elements
+        recipe = load_recipe("examples/recipe_python.yml")
+        recipe["datasets"].clear()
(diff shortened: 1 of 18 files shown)


bouweandela:
@lewisjared Is this the right place to start the `ref-metrics-esmvaltool` package? Or shall I put it in a separate repository under the [ESMValGroup](https://github.com/esMValGroup/) organization?

The tests could better, as could the functionality for editing recipes, running esmvaltool, and the CMEC output bundle, but maybe it's enough at this early stage.

Would it make sense to enable building the documentation for pull requests on readthedocs so they are easier to look at? Or are they available from the GitHub Actions and I just missed them?

packages/ref-metrics-esmvaltool/src/ref_metrics_esmvaltool/example.py
```diff
@@ -0,0 +1,108 @@
+from typing import Any
+
+import xarray
+from ref_core.datasets import FacetFilter, SourceDatasetType
+from ref_core.metrics import DataRequirement, Metric, MetricExecutionDefinition, MetricResult
+from ruamel.yaml import YAML
```
lewisjared, line 6: This is an aside, but why use ruamel over pyyaml?

packages/ref-metrics-esmvaltool/src/ref_metrics_esmvaltool/recipe.py
```diff
@@ -0,0 +1,209 @@
+from __future__ import annotations
+
+import subprocess
+from pathlib import Path
+from typing import TYPE_CHECKING, Any
+
+import pkg_resources
```
lewisjared, line 7: pkg_resources is deprecated. Is it possible to replace this usage with `importlib.resources`?

packages/ref-metrics-esmvaltool/src/ref_metrics_esmvaltool/recipe.py
```diff
@@ -0,0 +1,209 @@
...
+        "ensemble": "member_id",
+        "exp": "experiment_id",
+        "grid": "grid_label",
+        "mip": "table_id",
+        "short_name": "variable_id",
+    },
+}
+
+
+def as_isodate(timestamp: pd._libs.tslibs.timestamps.Timestamp) -> str:
```
lewisjared, line 30: Does the top-level `pd.Timestamp` typehint not work or `pd.core.api.Timestamp`?

lewisjared (commented):
Awesome stuff.

Can you add the tests to the CI. See https://github.com/CMIP-REF/cmip-ref/blob/c3ea68bbc9692056fdde122e872e61041d7da2dd/.github/workflows/ci.yaml#L96

lewisjared:
> @lewisjared Is this the right place to start the ref-metrics-esmvaltool package? Or shall I put it in a separate repository under the [ESMValGroup](https://github.com/esMValGroup/?rgh-link-date=2024-12-13T10%3A19%3A39Z) organization?

This is the right place for now. It can be refactored out in future, but I think it will help the development velocity by building the core and the metric packages in lockstep. It makes refactoring much easier.

> The tests and documentation could better, as could the functionality for editing recipes, running esmvaltool, and the CMEC output bundle, but maybe it's enough at this early stage.

Yeah definitely, but it is good to figure out the sticking points before we dive into adding additional helper functionality.

> Would it make sense to enable building the documentation for pull requests on readthedocs so they are easier to look at? Or are they available from the GitHub Actions and I just missed them?

The CI currently just verifies that the docs build in strict mode. I thought I clicked the button in RtDs to build on each MR, but that might not be true

packages/ref-metrics-esmvaltool/src/ref_metrics_esmvaltool/recipe.py
bouweandela: Thanks for flagging that. I will update it and opened an issue to report this to the authors of pooch: https://github.com/fatiando/pooch/issues/448.

bouweandela: Done in 5cdaa67

bouweandela:
> Can you add the tests to the CI. 

Done in c8f31e1. It would be nice to run those using the `make test` command, though I guess that is a bit tricky because of the code coverage collection settings.

packages/ref-metrics-esmvaltool/src/ref_metrics_esmvaltool/recipe.py
bouweandela: Changed in b3897fd. I'm not very familiar with `pandas`, so I just looked at the runtime type to figure out what it was.

packages/ref-metrics-esmvaltool/src/ref_metrics_esmvaltool/example.py
bouweandela: It supports roundtripping comments, which seems nice to have.

bouweandela:
> I thought I clicked the button in RtDs to build on each MR, but that might not be true

It seems to work now :partying_face: 

Would it be possible to also enable Codecov comments on pull requests? I can see that it is set up in the README, but it does not tell me which lines have missing coverage.

lewisjared:
> though I guess that is a bit tricky because of the code coverage collection settings.

Correct, otherwise I would have just used `make test`. There is a bit of duplication getting new packages setup

lewisjared:
> Would it be possible to also enable Codecov comments on pull requests? I can see that it is set up in the README, but it does not tell me which lines have missing coverage.

I'll have to look into it because it works for me (although I'm not making PRs from forks). The fork will mean that the `CODECOV_TOKEN` isn't set. I haven't had to support that before so any info would help

lewisjared:
Try merging main as I updated codecov to v5 (#44). That *should* work tokenless

bouweandela:
> Try merging main as I updated codecov to v5 (https://github.com/CMIP-REF/cmip-ref/pull/44). That should work tokenless

It works :partying_face:

bouweandela:
The documentation build failed because downloading the test data was too slow. Downloading several hundred megabytes of data to build the documentation may not be the fastest solution. Maybe we could build the documentation with smaller test data that we put in a GitHub repository?

lewisjared (approved):

PR #53, opened 2025-01-11
Prepend packages with cmip_

### LINKED ISSUE #22: PyPI package names
## The problem
We should start publishing to pypi to at bare minimum reserve our package naming, but to also ensure the publication workflow is 
fit for purpose.

I'd prefer to publish the following to PyPI:
* ref (Compute engine)
* ref-core (Shared functionality needed by metric providers)

Sadly, the `ref` pypi package is already claimed, but it has been unused for the last 6 years. I've sent an email to address associated with the project to see if they would be willing to assign the ownership of the ref PyPi to myself, but who knows if that will pan out.

We should make a backup plan. 

Perhaps we could publish as `cmip-ref` and `cmip-ref-core`. This results in a long name for the metrics packages (`cmip-ref-metrics-esmvaltool`). Not a major as these don't need to be published to pypi, but 🤷 

Another option would be to rename the `ref` package to something else (`ref-engine`?)

Any other ideas?

In #20 I've added a bump, release and deploy workflows for publishing new releases. 
This is untested for this repo, but we (@mikapfl  and I) have used this flow with success in other projects.

In addition, before publishing we need to resolve:
* [ ] #21 
* [x] #12 

## Definition of "done"
* [ ] Choose naming convention
* [ ] Rename packages if needed
* [ ] Bump a new package and deploy to pypi

lewisjared:
## Description

Updates the distribution and package names to be prepended by `cmip_`. 
The distribution name are also migrated to use `_` over `-`. 

## Checklist

- [ ] Tests added

Closes #22


lewisjared:
This doesn't rename the `packages/xxx` directories - that can happen in a follow up if its needed. I had some issues. With directories being recreated that I couldn't immediately diagnose so I moved on.

lewisjared:
@bouweandela apologies that I haven't given you the opportunity to review this, but I'm going to merge this now so that I can verify the release process. Hopefully this doesn't cause too many problems with your ECS MR.

PR #60, opened 2025-01-13
Add celery and a docker container

### LINKED ISSUE #78: Celery remote executor
## The problem
We have a basic local executor working, but we need to be able to run metrics out of process and asynchronously. 
Celery provides a useful interface to declare and run tasks out of process or even out of node to enable us to scale to support the amount of processing we required.

This could be a useful way of running tasks in production. I've had experience running multiple node celery deployments for moderately sized compute. I expect that the celery runner will be the default for production runs at least initially.

## Definition of "done"
* [x] Add a docker container for an example provider #60 
* [x] Add docker-compose for redis and example metric provider #60 
* [x] Support defining which executor/metric providers #77 
* [x] Add a way to scaffold out celery tasks
* [x] Add a celery executor (I expect this to require a bit of refactoring to support async tasks and to clarify the scope of an executor)

## Additional context

I'm not sure if we want to support one or many docker containers. Some teams already have an supported existing docker container so that might require minimal changes to prepare for the REF (install the core, metrics provider and celery package). I'd appreciate thoughts on this.

An alternative way to run tasks in production would be to run K8s tasks directly. Celery would be more resource intensive as it requires permanently allocated resources, but it can also be scale with usage.  K8s might provide the most control, but will be the hardest to build and debug. In theory the executor interface will support it, but I'm not as familiar with k8s.

lewisjared:
## Description

Adds an example docker container for the example metrics provider as well as the `cmip_ref_celery` package.
This docker container starts a worker process that registers a celery task for each metric. This worker process then waits for new tasks to be executed (not implemented) and then performs the metric execution.

This is the first of a few PRs to resolve #78 

## Checklist

- [x] Tests added


lewisjared:
I'm going to merge this as I've got a bunch of stuff to build on top of it and this isn't the most interesting bit

PR #63, opened 2025-01-16
Add ILAMB test integration

nocollier:
## Checklist

- [x] Tests added


nocollier:
@lewisjared The final failures here are due to `ilamb-run` not being found. I thought that the `pyproject.toml` for my package was to include only the requirements needed to make my glue work, excluding the actual ILAMB package. Have I misunderstood this? I could implement this glue in a few ways but have made it generate the needed setup files and then call a command which is available when ILAMB is installed.

This test won't pass until I can get a model with a continuous `tas` (and monthly) file that overlaps the time of the fluxnet2015 product. I have a CMEC bundle which is written into the ouput. I will just read it in and pass it out of the function once functional.

I need reference (observational) data for anything in ILAMB, but I can write some test that we can call if needed.

Once this works and we sort out how to stage my reference data, setting up all the ILAMB/IOMB metrics requests should be simple.

lewisjared:
Adding a dependency on mpi wouldn't be a show stopper especially given the target audience. It would suggest that we need at least a happy path to install via conda. We could use https://github.com/mpi4py/setup-mpi to setup mpi up on the test runner.

Lets discuss the tradeoff of ilamb vs ilamb3.

nocollier:
I am not sure what it means to add "a happy path to conda". ilamb2x is installable via conda and mpi4py if installed via conda does come with openmpi. I inserted the github actions you referenced in what seemed like the most reasonable location and this passes tests at least. 

Some thoughts about 2 vs 3:

#### ILAMB 2x
- Written almost 10 years ago and is based on netCDF4 and masked numpy arrays.
- I hand wrote data read and dimension recognition, which is now handled much more elegantly by xarray. This is a frequent issue as ilamb expands, I have to revamp this routine. It is tricky and error prone, especially given the wonky state of CMIP5 data which it needs to read. 
- For parallel logging, mpi4py is imported throughout making it necessary to install it even if running serially. I never designed the code to be used outside of this workflow and now it is what users want to do.
- It isn't slow, lazy loading was a thing way back in netCDF wrapper days. Parallelism is over the dataset/model pairs.
- However, most everything we have done is functional 
- I am trying to work as little as possible in this version and it becomes a moving target.

#### ILAMB 3
- Removes most ILAMB native objects in favor of community developed objects (xarray, pandas). I find that users do not understand objects and so I am trying to keep it as simple as I can.
- More of a library of functions to maximize interoperability and re-use.
- Focus on modular analysis, importable in notebooks as well as integrated in a workflow.
- Our general analysis gauntlet is not fully implemented yet as I continue working on design.

nocollier:
Closing this in favor of ilamb3 integration after a chat with @lewisjared

PR #64, opened 2025-01-16
Add temporal constraints

bouweandela:
## Checklist

- [ ] Tests added

### DIFF
--- a/packages/ref-core/src/cmip_ref_core/constraints.py
+++ b/packages/ref-core/src/cmip_ref_core/constraints.py
@@ -141,6 +141,74 @@ class RequireFacets:
 
 
 @frozen
+class RequireContiguousTimerange:
+    """
+    A constraint that requires datasets to have a contiguous timerange.
+    """
+
+    groupby: list[str]
+
+    def validate(self, group: pd.DataFrame) -> bool:
+        """
+        Check that all subgroups of the group have a contiguous timerange.
+        """
+        # Maximum allowed time difference between the end of one file and the
+        # start of the next file.
+        max_timedelta = pd.Timedelta(
+            days=31,  # Maximum number of days in a month.
+            hours=1,  # Allow for potential rounding errors.
+        )
+        group = group.dropna(subset=["start_time", "end_time"])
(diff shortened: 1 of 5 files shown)


packages/ref-core/src/cmip_ref_core/constraints.py
```diff
@@ -140,6 +140,82 @@ def validate(self, group: pd.DataFrame) -> bool:
...
+            start_series = sorted_group["start_time"]
+            end_series = sorted_group["end_time"]
+            # Sometimes the elements of start_series.values are of type datetime64[ns]
+            # and sometimes its elements are of type datetime.datetime.
+            # Convert both arrays to datetime.datetime objects to make sure they
+            # can be subtracted.
+            if hasattr(start_series, "dt"):
+                start_array = start_series.dt.to_pydatetime()
+            else:
+                start_array = start_series.values  # type: ignore[assignment]
```
bouweandela, line 182: @lewisjared I'm not very familiar with pandas, so I'm not sure how to generate a test case for this. However, this line is triggered by running `pytest packages/ref/tests/unit/test_solver.py::test_solve_metrics_default_solver`.

packages/ref-core/src/cmip_ref_core/constraints.py
```diff
@@ -140,6 +140,82 @@ def validate(self, group: pd.DataFrame) -> bool:
...
+            if hasattr(end_series, "dt"):
+                end_array = end_series.dt.to_pydatetime()
+            else:
+                end_array = end_series.values  # type: ignore[assignment]
+            diff = start_array[1:] - end_array[:-1]
+            contiguous = (diff < max_timedelta).all()
+            if not contiguous:
+                return False
```
lewisjared, line 190: Perhaps a debug log message here would be useful? We can always strip them out if it become noisy

lewisjared (approved):

packages/ref-core/src/cmip_ref_core/constraints.py
bouweandela: Added in 1eef0b9

PR #65, opened 2025-01-17
Add constraint for adding cell measures and ancillary variables

bouweandela:
## Description

Add a constraint for adding additional, supplementary datasets. Useful for e.g. adding cell measures and ancillary variables.

TODO:
- [x] Stop metric runs crashing
- [x] Add to example package metric
- [x] Add to ESMValTool ECS metric

Ideas for further improvement
- [x] Add a simpler API for cell measures and ancillary variables based on ancillary variable facets and project

## Checklist

- [x] Tests added


packages/ref-metrics-esmvaltool/src/cmip_ref_metrics_esmvaltool/metrics/example.py
```diff
@@ -24,10 +25,14 @@ class GlobalMeanTimeseries(ESMValToolMetric):
         DataRequirement(
             source_type=SourceDatasetType.CMIP6,
             filters=(FacetFilter(facets={"variable_id": ("tas",)}),),
-            # Add cell areas to the groups
-            # constraints=(AddCellAreas(),),
-            # Run the metric on each unique combination of model, variable, experiment, and variant
-            group_by=("source_id", "variable_id", "experiment_id", "variant_label"),
+            group_by=("instance_id",),
+            constraints=(
+                AddSupplementaryDataset(
```
nocollier, line 30: This is a great way to express this, going to use it in intake-esgf too.

packages/ref/src/cmip_ref/solver.py
```diff
@@ -52,13 +52,9 @@ def build_metric_execution_info(self) -> MetricExecutionDefinition:
...
             _subset = source_datasets[list(requirement.group_by)] if requirement.group_by else source_datasets
             unique_values = _subset.drop_duplicates()
-            if len(unique_values) != 1:
-                logger.error(f"Unique values: {unique_values}")
-                raise InvalidMetricException(
-                    self.metric,
-                    f"Expected a single group for {requirement.source_type} but got {len(unique_values)}",
-                )
-            key_values.extend(unique_values.iloc[0].to_list())
+            for i, row in enumerate(unique_values.itertuples(index=False), 1):
```
lewisjared, line 55: The name is included in file names so I'm not sure how this will scale on Windows which have a max length of 256, but let's see if that ever becomes an issue.

I guess we can't niavely join together the unique facets like `faceta_facetb1__facetb2_facet_c` because that breaks down once we have `faceta_facetb1__facetb2_facet_c1__facetc2`. Is that `faceta,facetb1,facetc1` and `faceta,facetb2,facetc2` or `faceta,facetb1,facetc2` and `faceta,facetb1,facetc2`...

lewisjared: I'm happy to try this for now, but I wonder if we should treat the datasets from the initial filter and groupby separate from `ancillary` data which are added via constraints. The ancillary data will still be used in the hash, but are somewhat separate. Thoughts?

lewisjared (approved):
Nice stuff

packages/ref/src/cmip_ref/solver.py
bouweandela: > The name is included in file names so I'm not sure how this will scale on Windows which have a max length of 256, but let's see if that ever becomes an issue.

As far as I can see, we are using a hash of the dataset to construct the filename: https://github.com/CMIP-REF/cmip-ref/blob/b58432cb2d6191f623c77cc4454c6ceabf98975b/packages/ref/src/cmip_ref/solver.py#L66

From the documentation here: https://github.com/CMIP-REF/cmip-ref/blob/b58432cb2d6191f623c77cc4454c6ceabf98975b/packages/ref-core/src/cmip_ref_core/metrics.py#L20-L26
it looks like you intended `key` to also a be a hash (maybe the same hash?). Would you like me to change that? The downside of a hash if that it is not human readable, but the advantage is that it doesn't grow and exceed limits like e.g. the path length or database column.

> I wonder if we should treat the datasets from the initial filter and groupby separate from ancillary data which are added via constraints. The ancillary data will still be used in the hash, but are somewhat separate. Thoughts?

This seems mostly useful to check that the grouping operation defined by the metric package developer makes sense and does not accidentally group different executions into the same group.

bouweandela: We could also compromise and use a combination of the identifier created by `group_by` and a hash for any remaining datasets.

lewisjared: >  we are using a hash of the dataframe with the filenames and their facets to construct the filename

Hmm I must have changed my mind at some point as I had an implementation where the output fragment was `pathlib.Path(self.provider.slug) / self.metric.slug / self.metric_dataset.name / self.metric_dataset.hash`, but I might have lost that change somewhere. I quite liked that because you get both a nice name to find the approximate location and then the innermost folder only has a few folders (one for each time the datasets change for that  group).

> If such a check would even be possible?

I don't think it would be possible without carrying around a more complex intermediate data representation.

lewisjared: > it looks like you intended key to also a be a hash (maybe the same hash?). Would you like me to change that?

I did intend it to be a somewhat sucinct, human-readable identifier, but still stable when new versions of datasets became available. It may be exposed to the frontend when we get to that. The new dataset will still have the same key, but the hash of the collection of datasets will differ because the non-groupby columns have changed (e.g. versions).

Otherwise we have to handle marking older executions as obsolete somehow. Happy to bounce some ideas around about how we can best track the latest "editions" of executions.

lewisjared: I think we are still a while away from production ready results so I don't mind merging and then revisiting how we achieve the functionality described above. The output naming scheme deserve some dedicated thought and justification about our decisions. We do have to have something we are happy with before we do any final runs because it will be be a pain to migrate since it will effect filenames and therefore likely URLs.

lewisjared: We don't track the `key` in the database layer, but eventually, we probably should. I could imagine it being part of a URL one day with a page to describe an execution. You would also be able to see that there are n previous versions of this execution.

bouweandela: I created #75 to keep track of this.

PR #68, opened 2025-01-22
chore: Update sample data to v0.3.0

lewisjared:
## Description

Updates the sample data  to v0.3.0 and streamlined some of the tests to minimise the number of hand rolled changes that are required when updating the sample data

## Checklist

- [x] Tests added


lewisjared:
@bouweandela I'm trying to make the tests that depend on the sample data use pytest-regressions where possible to make it easier to bump versions in future

Makefile
```diff
@@ -123,4 +123,4 @@ fetch-test-data:  ## Download any data needed by the test suite
 
 .PHONY: update-test-data-registry
 update-test-data-registry:  ## Update the test data registry
-	curl --output packages/ref/src/ref/datasets/sample_data.txt https://raw.githubusercontent.com/CMIP-REF/ref-sample-data/refs/heads/main/registry.txt
+	curl --output packages/ref/src/cmip_ref/datasets/sample_data.txt https://raw.githubusercontent.com/CMIP-REF/ref-sample-data/refs/heads/main/registry.txt
```
bouweandela, line 126: Maybe we could avoid the need to manually update the registry file and bump the version to the correct value: how about making this command bump the `SAMPLE_VERSION` variable to the latest release of the sample data and adding a bit of code in packages/ref/src/cmip_ref/testing.py that automatically downloads the file from `https://raw.githubusercontent.com/CMIP-REF/ref-sample-data/refs/tags/{SAMPLE_VERSION}/registry.txt` if it's not yet available locally? e.g. it could be stored in `pooch.os_cache("cmip_ref") / f"sample-data-registry-{SAMPLE_VERSION}.txt"`?

bouweandela (approved):

Makefile
lewisjared: We would also have to/should track the hash of the registry file. I quite like seeing the diff of the registry in the PR.  Maybe a future enhancement if we have to do this often.

bouweandela: > We would also have to/should track the hash of the registry file

The SAMPLE_VERSION git tag/commit already refers to a unique version of that file, so I don't think we would need another one.

If you like seeing the diff to make it more clear which files have changed, maybe we could just make the `make update-test-data-registry` also automatically figure out what the latest version is and update SAMPLE_VERSION, so we avoid some manual work and we are sure that the registry file always matches the SAMPLE_VERSION.

PR #72, opened 2025-01-22
PMP integration

lee1043:
## Description

PMP integration

## Checklist

- [ ] Tests added


lee1043:
@lewisjared while this is still in the early draft stage, please feel free to jump in if you find anything to improve.

packages/ref-metrics-pmp/pyproject.toml
```diff
@@ -0,0 +1,40 @@
...
+    "Programming Language :: Python :: 3.13",
+    "Topic :: Scientific/Engineering",
+]
+dependencies = [
+    "cmip_ref_core",
+    "pcmdi_metrics",
+]
+
+[tool.uv.sources]
+pmp = { git = "https://github.com/PCMDI/pcmdi_metrics" }
```
lewisjared, line 29:
```suggestion
pcmdi_metrics = { git = "https://github.com/PCMDI/pcmdi_metrics" }
```

This tells uv to install the pcmdi_metrics package from github. The name has to match to the dependencies section above and doesn't have to be the same as the python package name as long as it is the same above. Note that this will use the default branch instead of a particular tag.

You might need to run the following to regenerate the lockfile and your virtual environment.
```
uv lock
uv sync
```

packages/ref-metrics-pmp/tests/unit/test_metrics.py
```diff
@@ -0,0 +1,62 @@
...
+                "instance_id",
+            )
+        }
+    )
+
+
+def test_annual_mean(sample_data_dir, metric_dataset):
+    annual_mean = calculate_annual_mean_timeseries(metric_dataset["cmip6"].path.to_list())
+
+    assert annual_mean.time.size == 286
```
lewisjared, line 29:
```suggestion
    assert annual_mean.time.size == 11
```

This will also be needed for when you merge the main branch into this branch as we have a new set of sample data which don't include as many timesteps, but are contiguous in time.

lewisjared:
Looks good though! I was able to locally install and import the `pcmdi_metrics` package

packages/ref-metrics-pmp/pyproject.toml
lee1043: @lewisjared thank you for clarifying. I have a following question -- usual install practice for PMP has been through conda channel, which then installs pmp's dependencies. If REF installs packages from github repo, should I include all PMP's dependencies somewhere in REF as well?

lewisjared: Not for now. I'm perfectly happy to mock or comment out a required step until we are able to use the dependencies. We will solve that in the coming weeks because ESMValTool needs a similar install process. We will create a docker image that bundles all of the dependencies of the metrics packages

lewisjared:
I think you managed to find a bug in uv where it doesn't handle renaming the packages gracefully. Remove the lockfile (`uv.lock`) and recreate using `uv lock`

lewisjared (approved):

packages/ref-metrics-pmp/tests/unit/test_metrics.py
```diff
@@ -0,0 +1,62 @@
+import pathlib
+
+import pytest
+from cmip_ref_metrics_example.example import GlobalMeanTimeseries, calculate_annual_mean_timeseries
```
lewisjared, line 4:
```suggestion
from cmip_ref_metrics_pmp.example import GlobalMeanTimeseries, calculate_annual_mean_timeseries
```

packages/ref-metrics-pmp/tests/unit/test_metrics.py
```diff
@@ -0,0 +1,62 @@
+import pathlib
+
+import pytest
+from cmip_ref_metrics_pmp.example import GlobalMeanTimeseries, calculate_annual_mean_timeseries
```
lewisjared, line 4: Opps the import is called `AnnualCycle` instead of `GlobalMeanTimeseries`

packages/ref-metrics-pmp/tests/unit/test_metrics.py
```diff
@@ -0,0 +1,62 @@
...
+
+    assert annual_mean.time.size == 11
+
+
+def test_example_metric(tmp_path, metric_dataset, cmip6_data_catalog, mocker):
+    metric = GlobalMeanTimeseries()
+    ds = cmip6_data_catalog.groupby("instance_id").first()
+    output_directory = tmp_path / "output"
+
+    mock_calc = mocker.patch("cmip_ref_metrics_example.example.calculate_annual_mean_timeseries")
```
lewisjared, line 37: This will also need to be adjusted

packages/ref-metrics-pmp/tests/unit/test_metrics.py
```diff
@@ -0,0 +1,62 @@
...
+
+    mock_calc.return_value.attrs.__getitem__.return_value = "ABC"
+
+    definition = MetricExecutionDefinition(
+        output_directory=output_directory,
+        output_fragment=pathlib.Path(metric.slug),
+        key="global_mean_timeseries",
```
lewisjared, line 44: And this to match the expected metric name

lewisjared:
@lee1043 LGTM. Feel free to merge when you are ready

lee1043:
@lewisjared thanks for reviewing. I am merging it now while I will continue working on the actual metric calculation part in a following PR.

PR #73, opened 2025-01-22
missing integration stuff

nocollier:
## Description

Realized that I had missed some things needed for full integration.

## Checklist

- [x] Tests added

### DIFF
--- a/packages/ref-metrics-ilamb/src/cmip_ref_metrics_ilamb/example.py
+++ b/packages/ref-metrics-ilamb/src/cmip_ref_metrics_ilamb/example.py
@@ -71,8 +71,8 @@ class GlobalMeanTimeseries(Metric):
     Calculate the global mean timeseries for a dataset
     """
 
-    name = "Global Mean Timeseries"
-    slug = "global-mean-timeseries"
+    name = "ILAMB Global Mean Timeseries"
+    slug = "ilamb-global-mean-timeseries"
 
     data_requirements = (
         DataRequirement(
--- a/changelog/73.fix.md
+++ b/changelog/73.fix.md
@@ -1,0 +1,1 @@
+- Added overlooked code to fully integrate ilamb into ref.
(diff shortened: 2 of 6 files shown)


nocollier:
@lewisjared I discovered that I missed a good bit in the integration which meant tests ran but didn't include ilamb things. Now I have some errors, but these last ones I don't understand. Do you have any idea of what I have missed?

lewisjared:
Yup it turns out the tests are sensitive to new metrics and sample data so need to be tweaked each time either changes.  I've got a fix for this in a pr I'm about to merge so hopefully that will resolve this #68

lewisjared:
The changes look good though so feel free to assign it to me so I can merge once I fix the flakey test

packages/ref-metrics-ilamb/src/cmip_ref_metrics_ilamb/example.py
```diff
@@ -71,8 +71,8 @@ class GlobalMeanTimeseries(Metric):
     Calculate the global mean timeseries for a dataset
     """
 
-    name = "Global Mean Timeseries"
-    slug = "global-mean-timeseries"
+    name = "ILAMB Global Mean Timeseries"
```
lewisjared, line 74: I don't think that the slug had to be globally unique. The intention was that it is unique for a provider. The constraint is that the combination of the provider slug + the metric slug was unique

lewisjared:
Testing this locally after merging main looks to work
```
git checkout missing-ilamb
git merge origin/main
make virtual-environment
make test
```

packages/ref-metrics-ilamb/src/cmip_ref_metrics_ilamb/example.py
nocollier: I had a bunch (~10) failed tests due to some errors which reported 'unique slug' problems that were fixed when I made the change. But I will check this again.

Issue #80, opened 2025-01-30
Publish packages on conda-forge


bouweandela:
## The motivation

Some of the dependencies of metrics packages can only be installed from conda-forge. To facilitate an easy installation of both the REF and the metrics packages it depends on, it would be useful to publish the REF packages on conda-forge.

## The proposed solution

We create the recipes to publish the packages on conda-forge and open a pull request. Instructions are available here: https://conda-forge.org/docs/maintainer/adding_pkgs/

lewisjared:
1. We will be shuffling things around in the next week or so. Perhaps getting the packages conda installable might be out of sight for the hackathon. 

Perhaps as an alternative, we could vendor in an environment.yaml file for each package and a script to install all of the conda environments. This could be the "one-click" setup for now and would then be useful for building the provider-specific docker containers. Thoughts?

bouweandela:
Instead of a script, a `ref install` or similar command might be nicer.

bouweandela:
I had another go at this, but this pin
https://github.com/Climate-REF/climate-ref/blob/4d723324fb8b3e6a2fc306ad659808ec1c48c66f/packages/climate-ref-core/pyproject.toml#L56
makes it impossible at the moment because [version 1.0.5 of fastprogress is not available from conda-forge](https://anaconda.org/channels/conda-forge/packages/fastprogress/files).

lewisjared:
It might not be needed anymore as we've ripped out intake-esm

PR #81, opened 2025-01-31
Also build metrics packages

bouweandela:
## Description

In order to publish our packages through conda-forge #80, it would be useful to have all packages available on PyPI first. This adds the metrics packages to the build command. @lewisjared Is that sufficient to publish those on PyPI?

## Checklist

- [ ] Tests added

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,8 +45,16 @@ clean:  ## clean up temporary files
 build: clean  ## build the packages to be deployed to PyPI
 	cp LICENCE NOTICE packages/ref
 	cp LICENCE NOTICE packages/ref-core
+	cp LICENCE NOTICE packages/ref-metrics-example
+	cp LICENCE NOTICE packages/ref-metrics-esmvaltool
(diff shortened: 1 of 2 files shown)


lewisjared:
> @lewisjared Is that sufficient to publish those on PyPI?

From the repo side, I believe so. I've also registered each of the packages as a pending trusted publisher on pypi. On the next release this should cause the new packages to be published

lewisjared:
We can cut a new release before merging this PR and then I can manually build and publish the cmip_ref_metrics_* packages locally for that same tag. If we merge this and then tag, it would likely fail

lewisjared:
There are some additional checks that the built packages work as expected. There is a `scripts/test-install.py` which verifies that the package is importable without any extras. This is currently not doing what I wanted - it only checks the core and isn't flexible enough to be run for each package (I hoisted it out of a non-monorepo). I don' think that this needs to be fixed here, but just an FYI

lewisjared:
This has already landed

Issue #70, opened 2025-01-22
CMEC bundle generation and validation


lewisjared:
## The problem
The metrics provider packages currently hand-roll the generation of a CMEC-like output bundle. We already have multiple implementations of this (some of them are effectively noops because the data is hard coded). The functions to create an output bundle generally look like:

```
    def format_result(result_dir: Path) -> OutputBundle:
        """Format the result."""
        ecs_file = result_dir / "work/cmip6/ecs/ecs.nc"
        ecs = xarray.open_dataset(ecs_file)

        source_id = ecs.dataset.values[0].decode("utf-8")
        cmec_output = {
            "DIMENSIONS": {
                "dimensions": {
                    "source_id": {source_id: {}},
                    "region": {"global": {}},
                    "variable": {"ecs": {}},
                },
                "json_structure": [
                    "model",
                    "region",
                    "statistic",
                ],
            },
            # Is the schema tracked?
            "SCHEMA": {
                "name": "CMEC-REF",
                "package": "cmip_ref_metrics_esmvaltool",
                "version": __version__,
            },
            "RESULTS": {
                source_id: {"global": {"ecs": ecs.ecs.values[0]}},
            },
        }

        return cmec_output
```

That is they take some xarray-like result (either inmemory or serialised as an intermediate data product) and write out the nested dict that approximately conforms with CMEC.

We don't do any validation or verification of the resulting output bundles as of now. We don't even read them. We should check that the output from a metric execution results in a valid output bundle that the unified dashboard can read. 

@minxu74 @acordonez Do you have any pre-existing validation or generation routines for CMEC output?

We will likely end up generating scalar values, timeseries and gridded outputs. If we had to prioritise, I would focus on scalars and timeseries. 

## Definition of "done"
* [ ] Create a builder for the bundle to support scalars
* [ ] Create a builder for the bundle to support timeseries
* [ ] Validate the output bundle generated from a metric execution

## Additional context

If we need to add anything specific for the REF it should live in the `cmip_ref_core` package to be available for metrics packages

acordonez:
@lewisjared We don't have any validation routines for CMEC output bundles. In terms of generation I don't have any standardized code; it's always been built to fit the internals of an existing piece of metrics software.

In terms of writing metrics files that the unified dashboard can read, I think there might be a limit on the depth that metrics can be nested within the RESULTS object - something like 4 or 6 levels? @minxu74 would know.

lewisjared:
[...]

PR #84, opened 2025-02-03
Initially implement CMEC under cmip_ref_core

minxu74:
## Checklist

- [x] Tests added
- [x] Inline documents

Partially address the issue #70


packages/ref-core/src/cmip_ref_core/pycmec/tests/unit/test_cmec_metric.py
lewisjared, line 1: Can we hoist these tests into `packages/ref-core/tests/unit/cmec/...` rather than inlining them here

packages/ref-core/src/cmip_ref_core/pycmec/tests/unit/test_cmec_metric.py
```diff
@@ -0,0 +1,132 @@
...
+        "SCHEMA": {"name": "cmec", "version": "v1", "package": "ilamb v3"},
+        "DIMENSIONS": {
+            "json_structure": ["model", "metric"],
+            "model": {
+                "E3SM": {"name": "E3SM"},
+                "CESM2": {"name": "CESM2"},
+                "IPSL-CM5A-LR": {"name": "IPSL-CM5A-LR"},
+            },
+            "metric": {
+                "Ecosystem and Carbon Cycle": {"name": "Ecosystem and Carbon Cycle"},
```
lewisjared, line 22: The contents of the metrics aren't controlled, right? i.e. you don't have a CV for valid values for the `metric` dimension

The UD just takes the dimensions as provided

packages/ref-core/src/cmip_ref_core/pycmec/tests/unit/test_cmec_metric.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+@pytest.fixture
+def cmec_wrongdim_metric_dict(cmec_right_metric_dict):
+    cmec_right_metric_dict["DIMENSIONS"]["json_structure"] = ["model"]
+    return cmec_right_metric_dict
+
+
+def test_metric_right(cmec_right_metric_data):
+    try:
```
lewisjared, line 95: You don't need to wrap these in a try-except block as if the test raises an unhandled exception the test is marked as failed. You also get more information about the exception that is raised, where as this would swallow the exception

packages/ref-core/src/cmip_ref_core/pycmec/tests/unit/test_cmec_metric.py
```diff
@@ -0,0 +1,132 @@
...
+
+def test_dimen_right(cmec_right_dimen_data):
+    try:
+        MetricDimensions.model_validate(cmec_right_dimen_data)
+    except Exception as ex:
+        assert False, "test dimension object should be passed"
+
+
+def test_metric_wrongkm(cmec_wrongkw_metric_dict):
+    with pytest.raises(ValidationError):
```
lewisjared, line 109: This isn't clear which step should raise the `ValidationError`, is the the `model_validate` method call or the initialisation?

If they both should raise then they need to be wrapped in different `pytest.raises` blocks

packages/ref-core/src/cmip_ref_core/pycmec/metric.py
```diff
@@ -0,0 +1,244 @@
...
+    """
+    CMEC metric bundle object
+    """
+
+    model_config = ConfigDict(strict=True)
+
+    SCHEMA: Optional[MetricSchema] = None
+    DIMENSIONS: MetricDimensions
+    RESULTS: dict
+    PROVENANCE: Optional[dict] = None
```
lewisjared: We only support python >=3.10 so this can be written as:

```suggestion
    PROVENANCE: dict[str, Any] | None = None
```

Ruff will automatically rewrite these if you run `make ruff-fixes` or if you install the pre-commit hooks using `make virtual-environment` this will happen automatically when you commit.

The `dict` isn't a valid type hint by itself to mypy. It is a generic and needs to be provided with the type of the keys and values even if they are both Any.

packages/ref-core/src/cmip_ref_core/pycmec/output.py
```diff
@@ -0,0 +1,55 @@
...
+    """CMEC output bundle index object"""
+
+    index: str = "index.html"
+
+
+class OutputProvenance(BaseModel):
+    """CMEC output bundle provenance object"""
+
+    environment: dict  # Key/value pairs listing all relevant diagnostic and framework environment variables.
+    modeldata: list[str, ...]  # type: ignore # Path to the model data used in this analysis.
```
lewisjared, line 35: `list[str]` specifies that all items in the list are `str`s. `tuple[str, ...]` is the equiv for tuples.

```suggestion
    modeldata: list[str]  # type: ignore
    """Path to the model data used in this analysis."""
```

packages/ref-core/src/cmip_ref_core/pycmec/output.py
```diff
@@ -0,0 +1,55 @@
...
+class OutputIndex(BaseModel):
+    """CMEC output bundle index object"""
+
+    index: str = "index.html"
+
+
+class OutputProvenance(BaseModel):
+    """CMEC output bundle provenance object"""
+
+    environment: dict  # Key/value pairs listing all relevant diagnostic and framework environment variables.
```
lewisjared: The following will be more consistent with the rest of the project

```suggestion
    environment: dict[str, str]  
    """Key/value pairs listing all relevant diagnostic and framework environment variables."""
```

`dict` needs hints for the types of the keys and values. See another comment below

lewisjared (commented):
Here are an initial set of suggestions

lewisjared:
Ahh I see that you ran `ruff` over the code which resolved a bunch of the formatting suggestions

packages/ref-core/src/cmip_ref_core/pycmec/output.py
minxu74: will fix it.

packages/ref-core/src/cmip_ref_core/pycmec/tests/unit/test_cmec_metric.py
minxu74: Yes. I will move to that directory

packages/ref-core/src/cmip_ref_core/pycmec/tests/unit/test_cmec_metric.py
minxu74: Yes. they are not controlled by CMEC. UD just reads them and plots whatever metrics in the `metric` dimension.

packages/ref-core/src/cmip_ref_core/pycmec/tests/unit/test_cmec_output.py
```diff
@@ -0,0 +1,39 @@
...
+                "GlobalCarbon": {
+                    "version": "5.1",
+                    "title": "Global forest live biomass carbon",
+                },
+                "NBCD2000": {
+                    "version": "unknown",
+                    "name": "National Biomass and Carbon data set for the Year 2000",
+                },
+            },
+            "log": "cmec_output.log",
```
lewisjared, line 29: @nocollier As mentioned on slack the CMEC output bundle has a `log` field which takes a log file

packages/ref-core/src/cmip_ref_core/pycmec/tests/unit/test_cmec_metric.py
minxu74: Thanks, try, except are used in the tests that are expected success. I will remove them.

packages/ref-core/src/cmip_ref_core/pycmec/tests/unit/test_cmec_metric.py
minxu74: For pydantic, the "model_validate" is the same as class initialization as both methods will create the same instances and use the same validation code. They should fail or succeed together. But you are right. They are not clear, and I will separate them.

packages/ref-core/src/cmip_ref_core/pycmec/metric.py
minxu74: Thanks. yes. I have fixed these to meet the mypy requirement

minxu74:
@lewisjared I plan to add the changelog, so which one shall I use, "feature" or "improvement"? Thanks

lewisjared:
> @lewisjared I plan to add the changelog, so which one shall I use, "feature" or "improvement"? Thanks

This is a new `feature`

lewisjared:
Nice stuff @minxu74. I think all that remains is adding a dependency on pydantic to the `ref_core` package. 

```uv add --package cmip_ref_core pydantic```

File: packages/ref-core/src/cmip_ref_core/pycmec/metric.py [hunk 8]
```diff
@@ -0,0 +1,256 @@
+"""
+CMEC metric bundle class
```
lewisjared, line 2: It would be useful to have a link to the appropriate parts of the spec here

minxu74: Thanks, so, you mean to put the EMDS link within the docstring?

packages/ref-core/tests/unit/cmec/test_cmec_metric.py
```diff
@@ -0,0 +1,347 @@
...
+
+
+def test_gen_json_schema():
+    from cmip_ref_core.pycmec.metric import (
+        CMECGenerateJsonSchema,
+    )
+
+    cmec_model_schema = CMECMetric.model_json_schema(schema_generator=CMECGenerateJsonSchema)
+
+    assert cmec_model_schema == {
```
lewisjared, line 294: You might want to use pytest-regressions to track this schema as a separate file. I find that easier to maintain than inlining for things that are generated.

See the tests in packages/ref-core/tests/unit/test_datasets.py which use the `data_regression` and the docs here; https://pytest-regressions.readthedocs.io/en/latest/overview.html

packages/ref-core/src/cmip_ref_core/pycmec/metric.py
lewisjared: Yup

packages/ref-core/tests/unit/cmec/test_cmec_metric.py
minxu74: Cool. Thanks for the suggestion. I will follow it.

lewisjared:
Would you like to get this merged as is, and then do any follow up later. It is pretty self contained at the moment, but you would have to merge main back into this branch if you wanted to work on any other parts of the codebase as they have likely changed since you made this branch

minxu74:
@lewisjared I plan to submit another two commits tonight to address your suggestions and add more comments on the arguments of class methods. After that, if you and @acordonez agree, I can merge it to the main first as it does not interact with other ref components yet. We could create another PR to integrate the CMEC validators on other ref components.  I wonder if it works?

lewisjared:
I think that is a good idea. We are still actively building so I don't mind if this isn't 100% yet as it's more important to see how it works with the metrics providers

lewisjared (approved):

PR #94, opened 2025-02-09
Add ProposedMetricDefinition

lewisjared:
## Description

Adds `ProposedMetricDefinition` which represents the definition of a metric execution that is executor agnositic. Where possible paths are relative. This allows for executors to use different mount points for data.
## Checklist

- [x] Tests added


lewisjared:
I'm going to close this and make another PR to streamline as there is a few concepts lying around which were leading in this direction

PR #95, opened 2025-02-09
Draft: Add celery executor 

lewisjared:
## Checklist

- [ ] Tests added


lewisjared:
Need to rebase ontop of the replacement of #94 

We should also make a executor interface always be async. If you want to run a task synchrounously, locally you can just call the metric directly, or in the case of a more complex executor you can use a `.join` method to block until the async executions are complete.

lewisjared:
This is replaced by #104

PR #100, opened 2025-02-13
Use absolute paths everywhere

lewisjared:
## Description

This is an alternative to #94. Instead of tracking relative paths in the database we now track absolute paths.

## Checklist

- [x] Tests added


lewisjared:
I'm not going as far as adding a migration for any relative paths in the database. If anyone has issues with the existing relative paths I would recommend just deleting the db and starting again

PR #101, opened 2025-02-13
ILAMB relationship analysis

nocollier:
## Description

Adds ILAMB relationship analysis to the current metrics and flexibility to define new metrics via a yaml file.

## Checklist

- [x] Tests added


nocollier:
@lewisjared This is a big step forward in enabling the rest of the ILAMB analysis components that we will add. I had to skip a bunch of lines because I couldn't figure out how to test some of the behavior. See what you think.

packages/ref-metrics-ilamb/tests/unit/test_metrics.py
```diff
@@ -94,3 +95,50 @@ def test_standard_metric(tmp_path, cmip6_data_catalog):
...
+    assert output_bundle_path.exists()
+    assert output_bundle_path.is_file()
+
+
+def test_standard_fail():
+    try:
+        ILAMBStandard(registry_file="test.txt", sources={"gpp": "test/Grid/gpp.nc", "pr": "test/Grid/pr.nc"})
+        assert False
+    except ValueError:
+        pass
```
lewisjared, line 139:
```suggestion
    with pytest.raises(ValueError):
        ILAMBStandard(registry_file="test.txt", sources={"gpp": "test/Grid/gpp.nc", "pr": "test/Grid/pr.nc"})
```
Use pytest.raises to check that a given exception is raised. You can also assert on the exception message if you need to

packages/ref-metrics-ilamb/src/cmip_ref_metrics_ilamb/standard.py
```diff
@@ -143,43 +208,49 @@ def run(self, definition: MetricExecutionDefinition) -> MetricResult:
             csv_file = definition.to_output_path(f"{source_name}.csv")
             ref_file = definition.to_output_path("Reference.nc")
             com_file = definition.to_output_path(f"{source_name}.nc")
+            log_file = definition.to_output_path(f"{source_name}.log")
+            log_id = logger.add(log_file, backtrace=True, diagnose=True)
```
lewisjared, line 212: Nice soln to this. We could make this a context manager one day to make sure it is cleaned up and could also be used by the other packages

packages/ref-metrics-ilamb/src/cmip_ref_metrics_ilamb/standard.py
```diff
@@ -143,43 +208,49 @@ def run(self, definition: MetricExecutionDefinition) -> MetricResult:
...
+                dfs.to_csv(csv_file, index=False)
+                if not ref_file.is_file():  # pragma: no cover
+                    ds_ref.to_netcdf(ref_file)
+                ds_com[source_name].to_netcdf(com_file)
+                df_all.append(dfs)
             except Exception:  # pragma: no cover
-                raise AnalysisFailure(self.name, source_name)
-            df_scalars["source"] = df_scalars["source"].str.replace("Comparison", source_name)
-            dfs += [df_scalars]
+                logger.exception(f"ILAMB analysis {self.slug} failed for {source_name}.")
```
lewisjared, line 228: If no sources execute successful, should the execution should be deemed unsuccessful? If so the `MetricResult` returned from this func should `have the `successful` prop = False

lewisjared (approved):
I think some merge conflict resolution is needed, but LGTM

packages/ref-metrics-ilamb/tests/unit/test_metrics.py
nocollier: Thanks for the pointer, I am not terribly well versed in pytest.

packages/ref-metrics-ilamb/src/cmip_ref_metrics_ilamb/standard.py
nocollier: Ah yes, this is true. Will fix.

packages/ref-metrics-ilamb/src/cmip_ref_metrics_ilamb/standard.py
```diff
@@ -143,43 +208,49 @@ def run(self, definition: MetricExecutionDefinition) -> MetricResult:
...
             com_file = definition.to_output_path(f"{source_name}.nc")
+            log_file = definition.to_output_path(f"{source_name}.log")
+            log_id = logger.add(log_file, backtrace=True, diagnose=True)
 
-            # Run the analysis on the data/model pair
-            com = xr.open_mfdataset(sorted(grp["path"].to_list()))
             try:
-                df_scalars, ds_ref, ds_com[source_name] = run_analyses(reference_dataset, com, analyses)
+                # Load data and run comparison
+                ref = _load_reference_data(variable, self.registry, setup["sources"], setup["relationships"])
```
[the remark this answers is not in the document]
nocollier: I guess I ended up here without intending to. This also means that the ILAMB DataCollections do nothing. I don't want to push the REF structures down into ILAMB, but passing the registry isn't the way to couple. It is very convenient to just have ILAMB download what it needs as it needs it, but if an executor runs in parallel and multiple sources need the same data, we could have a race condition.

PR #104, opened 2025-02-17
Add a celery executor

lewisjared:
## Checklist

- [x] Tests added

## Description

* Make executors async by default
* Adds a celery executor


packages/ref-celery/src/cmip_ref_celery/executor.py
```diff
@@ -0,0 +1,126 @@
...
+
+        Parameters
+        ----------
+        provider
+            Provider for the metric
+        metric
+            Metric to run
+        definition
+            A description of the information needed for this execution of the metric
+            This include relative paths to the data files,
```
mikapfl, line 60:
```suggestion
            This includes relative paths to the data files,
```

packages/ref-celery/src/cmip_ref_celery/executor.py
```diff
@@ -0,0 +1,126 @@
...
+
+        try:
+            while results:
+                # Wait for a short time before checking for completed executions
+                time.sleep(refresh_time)
+
+                elapsed_time = time.time() - start_time
+
+                if elapsed_time > timeout:
+                    raise TimeoutError("All tasks did not complete within the specified timeout")
```
mikapfl, line 118:
```suggestion
                    raise TimeoutError("Not all tasks completed within the specified timeout")
```

packages/ref-celery/src/cmip_ref_celery/executor.py
```diff
@@ -0,0 +1,126 @@
...
+
+        Raises
+        ------
+        TimeoutError
+            If all executions aren't completed within the specified timeout
+        """
+        start_time = time.time()
+        refresh_time = 0.5  # Time to wait between checking for completed tasks in seconds
+
+        results = self._results.copy()
```
mikapfl, line 107: I don't know about the expected usage pattern, but if you call `run_metric` repeatedly, and every once in a while `join`, you'll leak ready results objects in `self._results`. For a steady-state thing where we run new metrics whenever needed and `join` every once in a while, the memory usage will grow without bounds. Is there a reason for this copy here? If you just don't call `.copy()` here, the leak goes away.

packages/ref-core/src/cmip_ref_core/executor.py
```diff
@@ -18,23 +20,62 @@ class Executor(Protocol):
...
         """
         Execute a metric
 
+        No results are returned from this method,
+        as the execution may be performed asynchrounusly so results may not be immediately available.
+
+        /// admonition | Note
+        In future, we may return a `Future` object that can be used to retrieve the result,
+        but that requires some additional work to implement.
+        ///
```
mikapfl, line 39: Depending on how we expect we'll run this eventually, it might also make sense to make `run_metric` async. But async python is of course a whole own world that we should only enter if we actually expect to use an event loop in the final daemon.

packages/ref/src/cmip_ref/executor/__init__.py
```diff
@@ -58,3 +66,76 @@ def import_executor_cls(fqn: str) -> type[Executor]:
...
+            metric_execution_result.output_fragment,
+            result.bundle_filename,
+        )
+
+        metric_execution_result.mark_successful(result.bundle_filename)
+
+        # TODO: This should check if the result is the most recent for the execution,
+        # if so then update the dirty fields
+        # i.e. if there are outstanding results don't make as clean
+        metric_execution_result.metric_execution.dirty = False
```
mikapfl, line 138: In an async world where new data comes in while we are processing the old data, handling this all via dirty fields by hand sounds harsh.

mikapfl (approved):

packages/ref/src/cmip_ref/executor/__init__.py
lewisjared: Yup I'm not super happy with this approach. The problem I'm trying to solve is the option to mark a execution as dirty to trigger it to be rerun on the next solve. It should stay dirty until a run started **after** being marked as dirty is complete.

mikapfl: maybe we just need dirty timestamps instead of booleans? "Dirty since"?

mikapfl: Or even dirty uuids?

packages/ref-core/src/cmip_ref_core/executor.py
mikapfl: happy to give an overview, maybe in a software meeting or something if it is also interesting for others.

packages/ref/src/cmip_ref/executor/__init__.py
lewisjared: So the dirty state would be cleared if the created datetime of a reslt is after the dirty datetime. I'm not quite sure if I follow the uuid thought. Probably worth an issue as it might be a good hackathon task

mikapfl: the uuid would be that when an execution starts, ideally atomically at the point when it collects input data, it also collects the dirty uuids that it will clean, and once we have the result, only if dirty uuid is still the same, it is cleared.

packages/ref-celery/src/cmip_ref_celery/executor.py
lewisjared: Good pickup. I intended to remove completed results from `self._results` upon completion. This is deliberately opaque to which results have been successfully as I don't want any consumers to rely on completed results until we have a proper solution

tests/integration/test_celery.py
```diff
@@ -0,0 +1,107 @@
...
+    # Run the solver which executes the metrics
+    solve_metrics(db_seeded, timeout=10, config=config)
+
+    results = db_seeded.session.query(MetricExecutionResult).all()
+    assert len(results)
+    for result in results:
+        assert result.successful
+        assert (config.paths.results / result.output_fragment / result.path).exists()
+
+    gc.collect()
```
lewisjared, line 107: @mikapfl This line took me hours. The `celery.AsyncResult`'s were being GCed after the redis container was torn down. This caused tests to hang sometimes depending on the order they were run. I went down a few different paths to try fix this before coalescing on this one line fix...

Celery testing is hard. We probably need an e2e test featuring docker containers etc, but that is a decent task in itself

mikapfl: That sounds pretty painful. Async always is, it seems, but tests hanging because garbage collection doesn't happen wouldn't be on my list of things to expect either.

Issue #67, opened 2025-01-22
Add support for Obs4MIPs as a source type


lewisjared:
## The problem
Currently we only support ingesting the CMIP6 datasets,
but metric evaluation requires observations.
The plan is to add obs4mips as another data source alongside the cmip6 data source.
This will allow for metrics to define a `DataRequirement` that references the obs4mips data.
The REF will then run the appropriate metrics executions depending on what observation data is available.

ILAMB uses its own preprocessed input data, which will be fetched as a different data source which is out of scope for this issue.

## Definition of "done"
* [ ]Add some obs4mips datasets to the test data (https://github.com/CMIP-REF/ref-sample-data/issues/10)
* [ ] Decide on the facets that are present in obs4mips datasets (Take a look at the CMIP6 implementation as an example)
* [ ] Add a new obs4mips module in `packages/ref/src/cmip_ref/datasets`
* [ ] Create a function that can extract those facets out of a local netCDF file
* [ ] Create a class which implements the `cmip_ref.datasets.base.DatasetAdapter` protocol
* [ ] Tests

## Additional context
The CMIP6 example uses `intake-esm` to do the facet extraction. I'm not 100% sure if we want to keep this dependency.

dhegedus99:
By dependency of the CMIP6 code on `intake-esm`, do you mean using the parser `ecgtools.parsers.parse_cmip6` at [L173](https://github.com/dhegedus99/cmip-ref/blob/59a3565b2a38c6023c586570c21daa249625df28/packages/ref/src/cmip_ref/datasets/cmip6.py#L173)?

lewisjared:
Correct. The intake-esm dependency can probably be replaced with a dependency on ecgtools as I don't think we use any of intake-esm anymore

lewisjared:
@dhegedus99 Could you create a draft PR if you have something almost ready?

PR #113, opened 2025-02-25
Obs4mips ingestion into REF

dhegedus99:
## Description
Added functionality to ingest obs4mips data by the REF, for issue #67.
## Checklist

- [ ] Tests added


packages/ref-core/tests/unit/test_datasets.py
```diff
@@ -11,6 +11,14 @@ def dataset_collection(cmip6_data_catalog) -> DatasetCollection:
     )
 
 
+@pytest.fixture
+def dataset_collection2(obs4mips_data_catalog) -> DatasetCollection:
```
lewisjared, line 15:
```suggestion
def dataset_collection_obs4mips(obs4mips_data_catalog) -> DatasetCollection:
```

This probably needs changes elsewhere, but is more explicity

packages/ref/src/cmip_ref/datasets/obs4mips.py
```diff
@@ -0,0 +1,344 @@
...
+        file_or_directory
+            File or directory containing the datasets
+
+        Returns
+        -------
+        :
+            Data catalog containing the metadata for the dataset
+        """
+        builder = Builder(
+            paths=[str(file_or_directory) + "/obs4MIPs"],
```
lewisjared, line 218: We shouldn't assume that the directory that is parsed is named obs4MIPs. Instead the user should pass the directory they actually want to ingest which might be a specific part of the directory tree

packages/ref/src/cmip_ref/datasets/obs4mips.py
```diff
@@ -0,0 +1,344 @@
...
+        """
+        self.validate_data_catalog(data_catalog_dataset)
+        unique_slugs = data_catalog_dataset[self.slug_column].unique()
+        if len(unique_slugs) != 1:
+            raise RefException(f"Found multiple datasets in the same directory: {unique_slugs}")
+        slug = unique_slugs[0]
+
+        dataset_metadata = data_catalog_dataset[list(self.dataset_specific_metadata)].iloc[0].to_dict()
+        dataset, created = db.get_or_create(self.dataset_cls, slug=slug, **dataset_metadata)
+        # if not created:
```
lewisjared, line 272: Why exclude this block?

packages/ref/src/cmip_ref/models/dataset.py
```diff
@@ -131,3 +131,70 @@ class CMIP6File(Base):
     """
 
     dataset = relationship("CMIP6Dataset", backref="files")
+
+
+class obs4MIPsDataset(Dataset):
```
lewisjared, line 136: Can we capitalise this class even if it differs from how obs4MIPs is normally capitalised

packages/ref/src/cmip_ref/models/dataset.py
```diff
@@ -131,3 +131,70 @@ class CMIP6File(Base):
...
+    version: Mapped[str] = mapped_column()
+
+    instance_id: Mapped[str] = mapped_column()
+    """
+    Unique identifier for the dataset.
+    """
+    __mapper_args__: ClassVar[Any] = {"polymorphic_identity": SourceDatasetType.obs4MIPs}  # type: ignore
+
+
+class obs4MIPsFile(Base):
```
lewisjared, line 172: Likewise for this class

packages/ref/src/cmip_ref/models/dataset.py
```diff
@@ -131,3 +131,70 @@ class CMIP6File(Base):
...
+    # File-specific metadata fields to track
+    end_time: Mapped[datetime.datetime] = mapped_column(nullable=True)
+    start_time: Mapped[datetime.datetime] = mapped_column(nullable=True)
+
+    path: Mapped[str] = mapped_column()
+    """
+    Prefix that describes where the dataset is stored relative to the data directory
+    """
+
+    dataset = relationship("obs4MIPsDataset", backref="files")
```
lewisjared, line 200:
```suggestion
    dataset = relationship("Obs4MIPsDataset", backref="files")
```

lewisjared (commented):
You will also need to check in the migrations as well

packages/ref/src/cmip_ref/datasets/obs4mips.py
```diff
@@ -0,0 +1,344 @@
...
+
+from cmip_ref.config import Config
+from cmip_ref.database import Database
+from cmip_ref.datasets.base import DatasetAdapter
+from cmip_ref.datasets.utils import validate_path
+from cmip_ref.models.dataset import Dataset, obs4MIPsDataset, obs4MIPsFile
+from cmip_ref_core.exceptions import RefException
+
+
+def _parse_datetime(dt_str: pd.Series[str]) -> pd.Series[datetime | Any]:
```
lewisjared, line 22: Is this the same as the CMIP implementation? If so we should import and use that until we need to diverge because obs4mips and CMIP6 are different

packages/ref/src/cmip_ref/datasets/obs4mips.py
```diff
@@ -0,0 +1,344 @@
...
+            )
+            or "v0"
+        )
+        return info
+
+    except Exception:
+        return {"INVALID_ASSET": file, "TRACEBACK": traceback.format_exc()}
+
+
+def _clean_branch_time(branch_time: pd.Series[str]) -> pd.Series[float]:
```
lewisjared, line 131: I don't think this is used

packages/ref/src/cmip_ref/models/dataset.py
```diff
@@ -131,3 +131,70 @@ class CMIP6File(Base):
...
+    """
+
+    __tablename__ = "obs4mips_dataset"
+    id: Mapped[int] = mapped_column(ForeignKey("dataset.id"), primary_key=True)
+
+    activity_id: Mapped[str] = mapped_column()
+    frequency: Mapped[str] = mapped_column()
+    grid: Mapped[str] = mapped_column()
+    grid_label: Mapped[str] = mapped_column()
+    init_year: Mapped[int] = mapped_column(nullable=True)
```
lewisjared, line 150: This value isn't extracted as far as I can tell

lewisjared:
@dhegedus99 Nice stuff. Add in the migrations and a test to ingest the obs4MIPs sample data and we are nearly there.

How confident are you that the list of metadata attributes is common across all datasets? We might need to be more defensive and set some of the database columns as `nullable=True`. Its pretty hard to figure that out with out downloading a bunch of data and just trying to see what breaks

packages/ref-core/tests/unit/test_datasets.py
dhegedus99: Newest commit should fix this.

packages/ref/src/cmip_ref/datasets/obs4mips.py
dhegedus99: Newest commit should fix this.

packages/ref/src/cmip_ref/models/dataset.py
dhegedus99: Newest commit should fix this.

packages/ref/src/cmip_ref/datasets/obs4mips.py
dhegedus99: Forgot to uncomment it, sorted now.

packages/ref/src/cmip_ref/datasets/obs4mips.py
dhegedus99: Yes it is, I've now deleted this function and imported the one from cmip6.

dhegedus99:
I've added the migrations, but there are 3 files (I think only one has actual content, but couldn't seem to delete the other two's migration  revision ids from the version history - if you can fix it that's great!
Regarding the facets, I've removed table_id as it seems to be from an earlier data specifications version for obs4mips and might not be used for newer datasets. I've also changed version to source_version_id which I think is the most consistent metadata used for recording the verison in obs4mips. Based on [ODS2.5](https://zenodo.org/records/11500474), I think all these facets are required attributes for new obs4mips datasets.

dhegedus99:
What other tests do you need?

packages/ref/src/cmip_ref/datasets/obs4mips.py
```diff
@@ -0,0 +1,311 @@
...
+                "source_id",
+                "source_type",
+                "variable_id",
+                "variant_label",
+            }
+        )
+    )
+
+    try:
+        with xr.open_dataset(file, chunks={}, use_cftime=True) as ds:
```
lewisjared, line 64:
```suggestion
	time_coder = xr.coders.CFDatetimeCoder(use_cftime=True)
        with xr.open_dataset(file, chunks={}, time_coder=time_coder) as ds:
```

use_cftime is deprecated (you might have saw warning messages when running the test suite

lewisjared:
> What other tests do you need?

There should be a unit test for the `parse_obs4mips` function as  that is quite standalone functionality and a perfect target for a unit test. There should be a test where you pass one of the sample data files and assert what results are returned. Another useful test would be to pass one of the CMIP6 files to this function and check that it returns the error result. 

<details>

<summary>Hint if you get stuck</summary>

```python
def test_parse_obs4mips(sample_data_dir):
    file_fragment = (
        pathlib.Path("obs4MIPs")
        / "NASA-JPL"
        / "AIRS-2-1"
        / "ta"
        / "gn"
        / "v20201110"
        / "ta_AIRS-2-1_gn_200209-201609.nc"
    )
    result = parse_obs4mips(str(sample_data_dir / file_fragment))

    assert result == {
        "project": "obs4MIPs",
        "institute": "NASA-JPL",
        ...
    }
```

</details>

     

If you want to get really fancy, you could parameterize the test so we can add more test cases easily

<details>

<summary>Hint if you get stuck</summary>

```python
@pytest.mark.parametrize(
    "file_fragment, exp",
    (
        (
            pathlib.Path("obs4MIPs")
            / "NASA-JPL"
            / "AIRS-2-1"
            / "ta"
            / "gn"
            / "v20201110"
            / "ta_AIRS-2-1_gn_200209-201609.nc",
            {
             "project": "obs4MIPs",
                ...
            },
        ),
    ),
)
def test_parse_obs4mips(sample_data_dir, file_fragment, exp):
    result = parse_obs4mips(str(sample_data_dir / file_fragment))

    assert result == exp

```

</details>

The existing tests are failing at the moment. Do you use an IDE such as PyCharm or VS Code? That makes it way easier to run a single test and see the results. If not you can run `uv run pytest packages/ref/tests/unit/datasets/test_obs4mips.py` to run all the test for obs4MIPs.

Looking at the current results there might be some inconsistencies in metadata names between CMIP6 and obs4MIPs. Have a go at cleaning that up and if you get stuck we can have a call

lewisjared:
We are aiming for 95% test coverage. If you run the test suite you will see the lines that aren't covered by any tests.

```
uv run --package cmip_ref \
		pytest packages/ref \
		-r a -v --doctest-modules --cov=packages/ref/src --cov-report=term --cov-append
```

dhegedus99:
I've added some tests. I realised the parse_obs4mips function also works with CMIP6 data because all obs4mips facets are also cmip6 facets (but the ecgtool.parser.parce_cmip6 doesn't work with obs4mips data) - with this in mind, I can either add an extra check to make sure parse_obs4mips throws an exception for cmip6 data, or just switch to using the same parsing function for reading in cmip6?

lewisjared:
It is best to raise an exception if a CMIP6 file gets passed to the obs4mips  parser. That indicates an incorrect path or source type

packages/ref/conftest.py
```diff
@@ -54,9 +55,35 @@ def db_seeded_template(tmp_path_session, cmip6_data_catalog) -> Path:
     return template_db_path
 
 
+@pytest.fixture(scope="session")
+def db_seeded_template_obs4mips(tmp_path_session, obs4mips_data_catalog) -> Path:
```
lewisjared, line 59: Can we just merge this with the above `db_seeded_template` unless you need a seeded db with just obs4mips data.

lewisjared (commented):
Looking good nearly there

packages/ref/src/cmip_ref/datasets/obs4mips.py
```diff
@@ -0,0 +1,308 @@
...
+            )
+            or "v0"
+        )
+        return info
+
+    except Exception:
+        return {"INVALID_ASSET": file, "TRACEBACK": traceback.format_exc()}
+
+
+class OBS4MIPSDatasetAdapter(DatasetAdapter):
```
lewisjared, line 107: Perhaps `Obs4MIPsDatasetAdapter`?

dhegedus99:
I think I've made all the changes you requested, so hopefully it's now ready to be a proper PR and merged. Let me know if there's anything else that needs added!

lewisjared (approved):
Nice stuff. Looks good to me. I added a changelog and the tests are all green so this is ready to merge

PR #117, opened 2025-02-28
Add management of conda environments

### LINKED ISSUE #111: Installing provider dependencies locally
## The problem
As an end-user, I want to be able to have a happy path to install the required dependencies for a package easily.

Currently for PMP and ESMValTool, there are conda environments that must exist outside of the project that must just exist. This is a somewhat opaque linkage. We should have someway of setting up a sane environment

This might be the basic setup and anything more advanced might have to be done manually.

Perhaps we have a CLI entry point for performing that action:

```
# Install a specific
ref providers install --provider cmip_ref_metrics_esmvaltool.provider
ref providers install --provider cmip_ref_metrics_ilamb.provider # This is a no-op because no additional deps are needed

# Or install all providers specified in the config
ref providers install --all
```

The `MetricProvider` class can be extended to include an `install` method that performs the actual installation.
The ESMValTool and PMP providers might need some provider-specific configuration to override the default installation location. This can be handled using our existing configuration format.

```
[[metric_providers]]
provider = "cmip_ref_metrics_pmp.provider"

[metric_providers.config]
environment_location = "/path/to/existing"
```
## Definition of "done"
* TBD

## Additional context
We will also bundle these dependencies into provider-specific docker containers

bouweandela:
## Description

Add a command to manage conda environments and allow metrics to run in them. 

Examples:

View all available providers and their conda environments
```bash
ref providers list
```

Install a conda environment for all providers that need them:
```bash
ref providers create-env
```

Install a conda environment for a particular provider
```bash
ref providers create-env --provider esmvaltool
```

A new configuration option is available to configure the path where the software will be installed:

```
[paths]
software = "/path/to/software"
```

It can be viewed with the command `ref config list`.

Note that running a metric from a provider that works with conda environments will cause automatic installation of the required environment.

Closes #111

## Checklist

- [x] Tests added


packages/ref-core/src/cmip_ref_core/metrics.py
```diff
@@ -345,3 +349,61 @@ def run(self, definition: MetricExecutionDefinition) -> MetricResult:
...
             The result of running the metric.
         """
+
+
+class CommandLineMetric(Metric):
+    """
+    Metric that can be run from the command line.
+    """
+
+    provider: Any  # TODO: CommandLineMetricsProvider causes circular import
```
lewisjared, line 359: Since it is just for typing this can be worked around by adding the following above

```
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from cmip_ref_core.providers import CommandLineMetricsProvider
```

I'm not 100% sure if the quotes are needed

```suggestion
    provider: "CommandLineMetricsProvider"
```

packages/ref-core/src/cmip_ref_core/providers.py
```diff
@@ -86,3 +99,176 @@ def get(self, slug: str) -> Metric:
...
+
+    @abstractmethod
+    def run(self, cmd: Iterable[str]) -> None:
+        """
+        Return the command to run.
+        """
+
+
+MICROMAMBA_EXE = (
+    "https://github.com/mamba-org/micromamba-releases/releases/latest/download/micromamba-linux-64"
```
lewisjared, line 164: Should we use the platform-independent installation method instead of assuming an architecture?

```
"${SHELL}" <(curl -L https://micro.mamba.pm/install.sh)
```

packages/ref-core/src/cmip_ref_core/providers.py
```diff
@@ -86,3 +99,176 @@ def get(self, slug: str) -> Metric:
...
+    """
+
+    def __init__(
+        self,
+        name: str,
+        version: str,
+        slug: str | None = None,
+    ) -> None:
+        super().__init__(name, version, slug)
+        self._conda_exe: Path | None = None
```
lewisjared, line 180: Should we check an environment variable if a particular system conda should be used and fallback to installing micromamba if needed?

It looks like this always installs micromamba. Is it a pretty safe assumption that micromamba will work in an HPC setting? I haven't used it before

An inspection of my env variables suggests that `CONDA_EXE` if conda is installed even if no base environment is activated. I do have to admit I don't use conda too much so this is somewhat outside my wheelhouse.

```suggestion
        self._conda_exe: Path | None = env.path("CONDA_EXE", default=None)
```

with an `from cmip_ref_core.env import env` above

bouweandela: I think it is more reliable to install our own. Outdated conda versions are a source of installation issues and the file is pretty small compared to the size of the conda environments.

bouweandela:
Thanks for reviewing! Note that this doesn't quite work yet, creating an environment works, but running the metric not yet. I hope I can get it working tomorrow morning

packages/ref-core/src/cmip_ref_core/metrics.py
bouweandela: Thanks! I couldn't get that to work before because then the interpreter complained that CommandLineMetricsProvider was not defined. Apparently, you need to either add quotes or use `from __future__ import annotations`.

packages/ref-core/src/cmip_ref_core/providers.py
bouweandela: Done in 7f3fbca

docs/development.md
```diff
@@ -65,6 +65,24 @@ for package in packages/cmip_ref_core packages/cmip_ref packages/cmip_ref_metric
...
+
+```bash
+ref --log-level=info providers create-env --provider ESMValTool
+```
+
+
+To update a conda-lock file, run for example:
+
+```bash
+uvx conda-lock -p linux-64 -f packages/ref-metrics-esmvaltool/src/cmip_ref_metrics_esmvaltool/requirements/environment.yml
```
lewisjared, line 82: Is there a reason to only use a single platform? Or can we add extra platforms?

Testing this out on my MacBook results in an empty env with a warning:
`warning  libmamba Selected packages for category 'main' are empty. The lockfile might not be resolved for your platform (osx-arm64).`

lewisjared: I'm assuming that it is because they don't have the appropriate dependencies available. For PMP osx-64 and linux-64 solve, but there is one dependency that doesn't have a distribution for win-64 or osx-arm64

lewisjared: And esmvaltool solves for all but win-64.

Bad luck if you want to run the REF natively on Windows. We probably need to surface that information somewhere

packages/ref-core/src/cmip_ref_core/providers.py
```diff
@@ -86,3 +116,233 @@ def get(self, slug: str) -> Metric:
...
+
+        with self.get_environment_file() as file:
+            cmd = [
+                f"{self.get_conda_exe(update=True)}",
+                "create",
+                "--yes",
+                "--file",
+                f"{file}",
+                "--prefix",
+                f"{self.env_path}",
```
lewisjared, line 322: I had to add "--platform osx-64" to support the pmp environment. Perhaps we need to add an optional platform configuration value in the Provider configuration to enable overriding the default platform?

docs/development.md
bouweandela: Indeed, ESMValTool doesn't work on Windows because not all dependencies are available for Windows. We usually recommend using WSL to people who are on Windows.

lewisjared:
Migrating PMP to use this was relatively straightforward: [`c5d25c5` (#127)](https://github.com/Climate-REF/climate-ref/pull/127/commits/c5d25c538f4646233e370c012e8074d4ec0caf9c)

I've done a clean ingest of the non-decimated sample data then solve and I've got successful results from all providers 🎉

packages/ref-core/src/cmip_ref_core/providers.py
bouweandela: Maybe pmp or one of its dependencies is not available on conda-forge for your default platform, see https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html for some background. Shall we try to tackle this in a follow up pull request? An alternative solution could be to make sure that all packages are available for all the platforms we support.

docs/development.md
bouweandela: Updated the command and lockfile to support osx in dadb44f

packages/ref-core/src/cmip_ref_core/providers.py
bouweandela: I looked into this a bit further: trying to solve the pcmdi_metrics environment for `arm64` results in `"nothing provides cdms2 needed by pcmdi_metrics-3.0.1-pyhd8ed1ab_0"`. The cdms package [is no longer maintained](https://github.com/CDAT/cdms) and it looks like @lee1043 is working on getting rid of the dependency https://github.com/PCMDI/pcmdi_metrics/issues/1203, so this problem will hopefully be solved soon.

bouweandela: Looking through the [`mamba` source code](https://github.com/mamba-org/mamba/blob/26785a6b50bea434cf8b7d964f92f0f407b34b0a/libmamba/src/api/configuration.cpp#L1372-L1380), it looks like the platform can be specified by using the environmental variables `CONDA_SUBDIR` or `MAMBA_PLATFORM`. Maybe that is a sufficiently good solution for now, since in the longer run this problem will be solved by removing the cdms dependency.

lewisjared: That will work. For Arm-based mac users, they will need to manually install the pmp environment via

`MAMBA_PLATFORM=osx-arm64 ref providers create-env --provider pmp`

I'll add that to the docs for the follow up PR for pmp

packages/ref-core/src/cmip_ref_core/providers.py
```diff
@@ -86,3 +116,233 @@ def get(self, slug: str) -> Metric:
...
+
+        cmd = [
+            f"{self.get_conda_exe(update=False)}",
+            "run",
+            "--prefix",
+            f"{self.env_path}",
+            *cmd,
+        ]
+        logger.info(f"Running {' '.join(cmd)}")
+        subprocess.run(cmd, check=True)  # noqa: S603
```
lewisjared, line 347: We might want to capture that subprocess output at some point. Although the executor might be a more consistent place to capture logs of the entire execution process

lewisjared (approved):

packages/ref-core/src/cmip_ref_core/providers.py
bouweandela: For ESMValTool, the logging messages are saved in its output directory. Stdout/stderr is lost though.

lewisjared: stdout/stderr comes through to the console when running an ESMValTool metric. We might want to suppress that unless we are at a debug log level as it is extremely noisy when running all providers

Everyone does it slightly different so we should try find common pattern that works for all. I see that you have two different log files. There is a field in the CMEC bundles for the log output, but we don't do anything with that field at the moment and logs aren't automatically copied to the "results" directory.

PR #123, opened 2025-03-06
Minor cleanups after using pycmec

lewisjared:
## Description

Added additional documentation
Added a create_template function to the CMECMetric class
## Checklist

- [x] Tests added


packages/ref-core/src/cmip_ref_core/pycmec/metric.py
minxu74, line 1: "SCHEMA" is not in the CMEC document. It is only used by the UD to determine the json format in the early stage of CMCE development. To avoid confusion, maybe we need to delete all the SCHEMA items.

minxu74:
@lewisjared Thanks a lot for fixing many typos and variable names that did not follow the name conventions. It looks great. My only concern is the "schema", I should not include it in the pycmec initially.

lewisjared:
> My only concern is the "schema", I should not include it in the pycmec initially.

I don't quite understand. Should "SCHEMA" not be in the CV?

minxu74:
> I don't quite understand. Should "SCHEMA" not be in the CV?

Yes. We need to delete it in the code and the sample json files if there is a "schema" in them

minxu74 (approved):

packages/ref-core/src/cmip_ref_core/pycmec/metric.py
lewisjared: Ok. Lets remove it. It has been in most metric bundles I have seen so this might touch a few of the metric providers.

It seems like a useful addition to the specification especially if we end up adopting a REF-specific addition to the EMDS.

PR #170, opened 2025-03-13
fix: Ensure that relative paths are always used

lewisjared:
## Checklist

- [x] Tests added

### DIFF
--- a/packages/ref-core/src/cmip_ref_core/metrics.py
+++ b/packages/ref-core/src/cmip_ref_core/metrics.py
@@ -17,6 +17,34 @@ if TYPE_CHECKING:
     from cmip_ref_core.providers import CommandLineMetricsProvider, MetricsProvider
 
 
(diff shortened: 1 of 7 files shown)


packages/ref-core/src/cmip_ref_core/metrics.py
```diff
@@ -17,6 +17,34 @@
     from cmip_ref_core.providers import CommandLineMetricsProvider, MetricsProvider
 
 
+def ensure_relative_path(path: pathlib.Path | str, root_directory: pathlib.Path) -> pathlib.Path:
+    """
+    Ensure that a path is relative to a root directory
+
+    If a path an absolute path, but not relative to the root directory, a ValueError is raised.
```
mikapfl, line 24:
```suggestion
    If a path is an absolute path, but not relative to the root directory, a ValueError is raised.
```

mikapfl (approved):

PR #174, opened 2025-03-13
Throws exception if no obs4mips-compliant datasets found during ingestion

### LINKED ISSUE #153: Missing error message when reading in obs data
This process fails without a useful error message when reading in a file that is not correctly formatted.

When trying to read in a netCDF observations file the script failed with a KeyError: 'start_time'. This was because the data file was unable to be read in because one or more of the obs4MIPs keys was not present. The user should be informed which key specifically is missing.

This error occurred for both HadISDH and GPCP files.
HadISDH: 
https://catalogue.ceda.ac.uk/uuid/1de7b50d827b4b4f966bd4e3ec5516ea/
hurs-blend_HadISDH_HadOBS_v1-5-0-2023f_19730101-20231231.nc
[link] https://dap.ceda.ac.uk/badc/ukmo-hadobs/data/insitu/MOHC/HadOBS/HadISDH-blend/mon/HadISDHTable/r1/v1-5-0-2023f/v20231231/hurs-blend/hurs-blend_HadISDH_HadOBS_v1-5-0-2023f_19730101-20231231.nc?download=1

GPCP:
The ESMValTool CMORised version.

## Failing Test
No unit test fail

## Expected behavior
We want an error messages that details why the file was not able to be read in.

## Screenshots
NA

## System

 - OS: linux RHEL9
 - Python version 3.12
 - Please also upload your `uv.lock` file (first run `uv lock` to make sure the lock file is up-to-date)

## Additional context
The error is within climate-ref/packages/ref/src/cmip_ref/datasets/obs4mips.py.
The file we tried did not have the correct keys as listed in the parse_obs4mips() function.

logger.execution() error messages need to be added lines 108-111 there the TypeError and Exception are raised.

dhegedus99:
These changes resolve #153. Warnings are logged when the REF is trying to ingest non-obs4mips compliant data, and if there were no obs4mips-compliant datasets ingested it throws an exception.

## Checklist

- [ ] Tests added

### DIFF
--- a/packages/ref/src/cmip_ref/datasets/obs4mips.py
+++ b/packages/ref/src/cmip_ref/datasets/obs4mips.py
@@ -109,8 +109,11 @@ def parse_obs4mips(file: str) -> dict[str, Any | None]:
         )
         return info
 
-    except (TypeError, AttributeError):
-        logger.warning(traceback_message)
+    except (TypeError, AttributeError) as err:
+        if (len(err.args)) == 1:
+            logger.warning(str(err.args[0]))
+        else:
+            logger.warning(str(err.args))
         return {"INVALID_ASSET": file, "TRACEBACK": traceback_message}
     except Exception:
         logger.warning(traceback.format_exc())
--- a/packages/ref/src/cmip_ref/datasets/obs4mips.py
+++ b/packages/ref/src/cmip_ref/datasets/obs4mips.py
@@ -109,7 +109,7 @@ def parse_obs4mips(file: str) -> dict[str, Any | None]:
         )
         return info
 
-    except TypeError, AttributeError:
+    except (TypeError, AttributeError):
         logger.warning(traceback_message)
         return {"INVALID_ASSET": file, "TRACEBACK": traceback_message}
     except Exception:
--- a/packages/ref/src/cmip_ref/datasets/obs4mips.py
+++ b/packages/ref/src/cmip_ref/datasets/obs4mips.py
@@ -109,10 +109,7 @@ def parse_obs4mips(file: str) -> dict[str, Any | None]:
         )
         return info
 
-    except TypeError:
-        logger.warning(traceback_message)
-        return {"INVALID_ASSET": file, "TRACEBACK": traceback_message}
-    except AttributeError:
+    except TypeError, AttributeError:
         logger.warning(traceback_message)
         return {"INVALID_ASSET": file, "TRACEBACK": traceback_message}
     except Exception:
--- a/packages/ref/src/cmip_ref/datasets/obs4mips.py
+++ b/packages/ref/src/cmip_ref/datasets/obs4mips.py
@@ -63,6 +63,11 @@ def parse_obs4mips(file: str) -> dict[str, Any | None]:
     try:
         time_coder = xr.coders.CFDatetimeCoder(use_cftime=True)
         with xr.open_dataset(file, chunks={}, decode_times=time_coder) as ds:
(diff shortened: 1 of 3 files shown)


dhegedus99:
I've created a changelog item on my branch that this pull request is for, but it hasn't shown up here in the PR itself for some reason. Do I need to create the changelog file in Climate-REF:main instead?

packages/ref/src/cmip_ref/datasets/obs4mips.py
```diff
@@ -106,8 +110,13 @@ def parse_obs4mips(file: str) -> dict[str, Any | None]:
         return info
 
     except TypeError:
+        logger.warning(traceback_message)
+        return {"INVALID_ASSET": file, "TRACEBACK": traceback_message}
+    except AttributeError:
+        logger.warning(traceback_message)
```
lewisjared, line 116: You can catch multiple exceptions in the sample block

```suggestion
    except TypeError, AttributeError:
        logger.warning(traceback_message)
```

lewisjared:
Odd the changelog appeared once you added that extra commit, because I was halfway through asking you to add a changelog.

Turns out you need to wrap the exceptions in parentheses

packages/ref/src/cmip_ref/datasets/obs4mips.py
```diff
@@ -105,9 +109,11 @@ def parse_obs4mips(file: str) -> dict[str, Any | None]:
         )
         return info
 
-    except TypeError:
+    except TypeError, AttributeError:
```
lewisjared, line 112:
```suggestion
    except (TypeError, AttributeError):
```

lewisjared:
This already has test coverage from elsewhere apparently, but it would be good to add a test that if you try ingest an empty folder it raises an exception. That test in particular, would be useful if we ever want to change that behaviour (and its a pretty easy test to add).

Other than that looks goo

packages/ref/src/cmip_ref/datasets/obs4mips.py
```diff
@@ -105,9 +109,11 @@ def parse_obs4mips(file: str) -> dict[str, Any | None]:
         )
         return info
 
-    except TypeError:
+    except (TypeError, AttributeError):
+        logger.warning(traceback_message)
```
mikapfl, line 113: Because you are re-using Python built-in exceptions (`TypeError` and `AttributeError`), they can also be raised by other things than you explicitly raising them, and then the `traceback_message` variable would not be set, which will raise another `NameError`, which will be a bit confusing. This problem also existed previously with only the `TypeError`, so obviously not super serious, but maybe worth to fix.

One way to deal with this is to instead use the `args` attribute that `TypeError` and `AttributeError` always have:
```suggestion
    except (TypeError, AttributeError) as err:
        logger.warning(str(err.args))
```
Note that it will look a bit different because a `traceback_message = "example"` will lead to `str(err.args) = '("example",)'`. If you want to avoid that, you have to do some dance with `if(len(err.args)) == 1: str(err.args[0]) else str(err.args)` or so because if this is an error from something else you might not have args at all, or more than one arg.

mikapfl (approved):
Looks good to me, thanks. Only found a small inconsistency, which already existed in the changed code, but now is more likely to trigger.

packages/ref/src/cmip_ref/datasets/obs4mips.py
dhegedus99: resolved in my latest commit f02537c

PR #183, opened 2025-03-17
feat: Add basic support for a controlled vocabulary

lewisjared:
## Description

Adds the basic framework for a controlled vocabulary for metric values and their dimensions. 

This is still a prototype, but it provides an example about how we could capture information about the dimensions we will use in the metric bundles.

The CMIP cvs are generally JSON files, one for each controlled dimension. They capture additional metadata about each of the controlled values. We need to support the definition of all permitted dimensions in addition to the controlled ones which has led to more structured schema as to also capture that information. Yaml was a more readable alternative imho 

If the cv grows to where it is difficult to follow we can split it out into different dimensions as well. The ref needs different information compared to the CMIP cvs so they couldn't be used directly, but may draw from them in a more complete version of the cv

## Checklist

- [x] Tests added

### DIFF
--- a/changelog/183.feature.md
+++ b/changelog/183.feature.md
@@ -1,0 +1,4 @@
+Add the basic framework for enforcing a controlled vocabulary
+for the results in a CMEC metrics bundle.
+This is still in the prototype stage
+and is not yet integrated into post-metric execution processing.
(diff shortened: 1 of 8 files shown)


minxu74:
@lewisjared very nice implementation. Now we could validate `DIMENSIONS` and then further validate `RESULTS` using the validated dimensions. Maybe we could add a method to generate the `DIMENSIONS` dictionary just using the json_structure and predefined the CVs. Thanks.

minxu74 (approved):

PR #187, opened 2025-03-24
test: Add an integration test for the AR7 FT metric providers

lewisjared:
## Description

Add an integration test for the AR7 FT setup. This adds a workflow which is run nightly.

The results from the test suite are now archived to enable debugging if the suite fails

PMP is currently failing due to missing data so is ignored.

## Checklist

- [x] Tests added

### DIFF
--- a/packages/ref-metrics-ilamb/src/cmip_ref_metrics_ilamb/standard.py
+++ b/packages/ref-metrics-ilamb/src/cmip_ref_metrics_ilamb/standard.py
@@ -111,10 +111,7 @@ def _load_csv_and_merge(output_directory: Path) -> pd.DataFrame:
     Load individual csv scalar data and merge into a dataframe.
     """
     df = pd.concat(
-        [
-            pd.read_csv(output_directory / f, keep_default_na=False, na_values=["NaN"])
-            for f in output_directory.glob("*.csv")
-        ]
+        [pd.read_csv(f, keep_default_na=False, na_values=["NaN"]) for f in output_directory.glob("*.csv")]
     ).drop_duplicates(subset=["source", "region", "analysis", "name"])
     df["name"] = df["name"] + " [" + df["units"] + "]"
     return df
--- a/tests/integration/test_ar7_ft.py
+++ b/tests/integration/test_ar7_ft.py
@@ -72,7 +72,8 @@ def test_solve_ar7_ft(
     print(df)
 
     # Check that all 3 metric providers have been used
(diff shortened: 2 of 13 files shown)


packages/ref-metrics-ilamb/src/cmip_ref_metrics_ilamb/standard.py
```diff
@@ -111,10 +111,7 @@ def _load_csv_and_merge(output_directory: Path) -> pd.DataFrame:
     Load individual csv scalar data and merge into a dataframe.
     """
     df = pd.concat(
-        [
-            pd.read_csv(output_directory / f, keep_default_na=False, na_values=["NaN"])
-            for f in output_directory.glob("*.csv")
-        ]
+        [pd.read_csv(f, keep_default_na=False, na_values=["NaN"]) for f in output_directory.glob("*.csv")]
```
lewisjared, line 114: @nocollier FYI: output_directory.glob returns a path that includes the output directory

tests/integration/test_ar7_ft.py
```diff
@@ -0,0 +1,79 @@
...
+    # Solve
+    # This will also create conda environments for the metric providers
+    invoke_cli(["--verbose", "solve", "--timeout", f"{60 * 60}"])
+
+    executions = db.session.query(MetricExecution).all()
+    df = create_execution_dataframe(executions)
+    print(df)
+
+    # Check that all 3 metric providers have been used
+    # TODO: Update once the PMP metrics are solving
```
lewisjared, line 75: @lee1043 FYI the integration test doesn't currently run any PMP metrics due to the missing obs4MIPs data. Do you have temporary URLs for these obs data so that we can add them to the sample data?

lee1043: obs4MIPs-complying reference datasets for PMP's modes of variability metrics are available via the following links:

- HadISST: https://pcmdiweb.llnl.gov/pss/pmpdata/obs4MIPs_PCMDI_monthly/MOHC/HadISST-1-1/mon/ts/gn/v20210727/ts_mon_HadISST-1-1_PCMDI_gn_187001-201907.nc

- NOAA-20CR: https://pcmdiweb.llnl.gov/pss/pmpdata/obs4MIPs_PCMDI_monthly/NOAA-ESRL-PSD/20CR/mon/psl/gn/v20210727/psl_mon_20CR_PCMDI_gn_187101-201212.nc

Could you add them to the workflow so PMP can be included for the test?

PR #203, opened 2025-03-26
Nicer error message when specifying wrong columns in `ref datasets list`

bouweandela:
## Description

Adds a nicer error message when running `ref datasets list` with the `--column` argument. If a column is specified that is not available in the dataframe, a list of available columns will be shown.

## Checklist

- [x] Tests added

### DIFF
--- a/packages/ref/src/cmip_ref/cli/datasets.py
+++ b/packages/ref/src/cmip_ref/cli/datasets.py
@@ -4,6 +4,7 @@ View and ingest input datasets
 
 import errno
 import os
+from collections.abc import Iterable
 from pathlib import Path
 from typing import Annotated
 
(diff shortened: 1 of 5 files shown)


packages/ref/src/cmip_ref/cli/datasets.py
```diff
@@ -43,8 +44,17 @@ def list_(
...
+        if missing:
+
+            def format(columns: Iterable[str]) -> str:
+                return ", ".join(f"'{c}'" for c in sorted(columns))
+
+            logger.error(
+                f"Column{'s' if len(missing) > 1 else ''} "
+                f"{format(missing)} not found in data catalog. "
+                f"Choose from: {format(data_catalog.columns)}"
+            )
```
mikapfl, line 57:
```suggestion
            def _format(columns: Iterable[str]) -> str:
                return ", ".join(f"'{c}'" for c in sorted(columns))

            logger.error(
                f"Column{'s' if len(missing) > 1 else ''} "
                f"{_format(missing)} not found in data catalog. "
                f"Choose from: {_format(data_catalog.columns)}"
            )
```

mikapfl (approved):
Nice, just a small paranoid thing from me: `format` is a python built-in, so better to use a different name, like `_format`, otherwise there might be surprises in the same namespace later.

packages/ref/src/cmip_ref/cli/datasets.py
bouweandela: Well spotted! I fixed it and enabled the ruff rule that warns about this: https://docs.astral.sh/ruff/rules/#flake8-builtins-a in a6b365a.

PR #208, opened 2025-03-26
Add TCRE metric

### LINKED ISSUE #193: Transient Climate Response to cumulative emissions of carbon dioxide

bouweandela:
## Description

Added Transient Climate Response to Cumulative CO2 Emissions (TCRE) metric to the ESMValTool metrics package.

Sample data in https://github.com/Climate-REF/ref-sample-data/pull/22. Note that this appears to be the only model in CMIP6 that provides the required variables in the required experiments.

Closes #193 

## Checklist

- [x] Tests added


lewisjared:
> Note that this appears to be the only model in CMIP6 that provides the required variables in the required experiments.

This is more of a science question, but I believe it is possible (and useful) to calculate TCRE from non CO2-emissions driven runs. Is that out of scope for the REF?  Are we just focussing on emissions driven runs? I assume the TCRE from the other 10 models in AR6 were concentration-driven (I haven't dived into the details).

The FT contains a number of flat10 experiments which are designed to diagnose ZEC and TCRE so the CMIP7 FT version of this metric should leverage them as they are new science. I'm sure @benmsanderson would be very interested in us using those experiments where possible. 

@benmsanderson If we were to report TCRE values for different models in the AR7 FT should we calculate via the 1pctCO2 experiment *and/or* the flat10 experiment? Can we mix results calculated from the different experiments if they don't have flat10 experiments?

We will not resolve that in this PR so I don't think it blocks this from landing. The flat10 experiments might be a post-beta addition.

CC @znicholls

benmsanderson:
In the paper we found that TCRE derived from 1pctCO2 is about ~0.1K higher than from flat10 in CMIP models - broadly due to the fact that compatible emissions are weighted towards the end of the 70year period and you get a transient pulse warming in most models.  So - I wouldn't mix them if possible, but there's a reasonable argument that the flat10 number is more relevant to reality -  i.e. recent history looks more like flat10 than it does like 1pctCO2 (which has emissions 4x present day levels in some cases towards the end of the 70 year period).

In short - report the TCRE from flat10 and 1pctCO2 as indepedent quantities
and *definitely* report ZEC from the two experiments as indpendent, they are very different.

bouweandela:
@schlunma @bettina-gier Since you implemented TCRE and ZEC in ESMValTool respectively, could you please comment here?

znichollscr:
> In short - report the TCRE from flat10 and 1pctCO2 as indepedent quantities
> and _definitely_ report ZEC from the two experiments as independent, they are very different

Big +1 for this from me. (With the awkward implication that you will need a new metric calculation to do this from flat10, and also a new recipe to do it from 1pctCO2 because you need to do it from a concentration-driven 1pctCO2 run and back out the inverse emissions rather than relying on reported emissions for exactly the reason you've already noted: most models don't report the data required to do it the way that is implemented here. Probably one for a follow up PR, although naming will be fun with three different ways to calculate the same thing.)

schlunma:
As already mentioned [here](https://github.com/ESMValGroup/ESMValTool/pull/3904#issue-2827991338), the only reason we use the `esm-1pctCO2` runs in the [example recipe](https://github.com/ESMValGroup/ESMValTool/blob/main/esmvaltool/recipes/recipe_tcre.yml) is that the `esm-flat10` experiment is not available for CMIP6.

Our idea was to diagnose TCRE only from the `esm-flat10` runs in the REF as described in Ben's paper. It would certainly be possible to also report TCRE for the `1pctCO2` runs, and I fully agree that we shouldn't mix those. This would probably require slight changes in the recipe and/or diagnostic code within ESMValTool.

packages/ref-metrics-esmvaltool/src/cmip_ref_metrics_esmvaltool/metrics/tcre.py
```diff
@@ -0,0 +1,125 @@
...
+            },
+            "fco2antt": {
+                "preprocessor": "global_cumulative_sum",
+                "additional_datasets": [fco2antt_esm_1pctCO2],
+            },
+        }
+        recipe["diagnostics"].pop("barplot")
+
+    @staticmethod
+    def format_result(result_dir: Path) -> OutputBundle:
```
lewisjared, line 103: We now have concrete types `cmip_ref_core.pycmec.output.CMECOutput` and `cmip_ref_core.pycmec.output.CMECMetric` instead of this loose type hint.
The content here is also representative of a metrics bundle, not an output bundle. 

The output bundle contains references to plots and datafiles, where as the metric bundle contains the calculated scalars. The output bundle content is not currently implemented for ESMValTool (see `packages/ref-metrics-pmp/src/cmip_ref_metrics_pmp/pmp_driver.py` for an example of including data and plots in an output bundle).

I'd recommend fixing this up in a follow-up PR as is will impact all of the ESMValTool metrics

PR #209, opened 2025-03-27
Match all facet values when using `FacetFilter` with `keep=False`

bouweandela:
## Description

For the TCRE metric #208, I need variables `"tas"` and `"fco2antt"` from the `"esm-1pctCO2"` experiment and variable `"tas"` from the `"esm-piControl"` experiment. Therefore I wrote the following filters:
```
FacetFilter(
    facets={
        "variable_id": ("tas", "fco2antt"),
        "frequency": "mon",
        "experiment_id": ("esm-1pctCO2", "esm-piControl"),
    },
),
FacetFilter(
    facets={
        "variable_id": "fco2antt",
        "experiment_id": "esm-piControl",
    },
    keep=False,
),
```
The first filter selects all files with `"variable_id"` `"tas"` and `"fco2antt"` from both experiments and the second filter then excludes files with `"variable_id"` `"fco2antt"` from the `"esm-piControl"` experiment.

The implementation here makes the above filters work as expected.

## Checklist

- [x] Tests added

### DIFF
--- a/changelog/209.fix.md
+++ b/changelog/209.fix.md
@@ -1,0 +1,2 @@
+Fixed the behaviour of FacetFilter with `keep=False` so all facets need to match
+before excluding a file.
--- a/packages/ref-core/src/cmip_ref_core/metrics.py
+++ b/packages/ref-core/src/cmip_ref_core/metrics.py
@@ -328,6 +328,7 @@ class DataRequirement:
             Filtered data catalog
         """
         for facet_filter in self.filters:
+            mask = None
             for facet, value in facet_filter.facets.items():
                 clean_value = value if isinstance(value, tuple) else (value,)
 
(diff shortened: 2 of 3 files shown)


mikapfl (approved):
Lgtm, thanks.

It still feels to me like we should be able to just use `pandas_indexing`s filters instead of discovering problems like these on our own, but that's a topic for another day.

bouweandela:
I'm not very familiar with pandas, could you explain that in a bit more detail? I had another look at the [pandas indexing documentation](https://pandas.pydata.org/docs/user_guide/indexing.html) and found a simpler solution that uses a bit more of pandas in 452abf5.

mikapfl:
Ah, sorry, I was talking about https://pandas-indexing.readthedocs.io/en/latest/ , where we can write stuff like
```python
filter = isin(variable_id=["tas", "fco2antt"], frequency="mon", experiment_id=["esm-1pctCO2", "esm-piControl"]) & ~isin(variable="fco2antt", experiment_id="esm-piControl")
data_catalog = data_catalog.loc[filter]
```
which I think is at least equally readable compared to our custom FacetFilter class and syntax, but someone else already built it for us. However, it is of course another dependency, and we already have FacetFilters and they work and we don't want to refactor everything. So, sorry for the distraction.

PR #227, opened 2025-04-02
Support fetching annual cycle data

lewisjared:
## Description

Unifies the data fetching across the packages enabling providers to register any additional data registries. These data registries can then be fetched using `ref datasets fetch-data`, either to pre-cache the data or to write the data to a given directory so they can be ingested.

 Available registries: 
 * obs4ref - Full set of obs datasets that will be ingested into the REF for the beta release. These will eventually be published to obs4MIPs
 * sample-data - Sample data (Needed by test suite)
 * ilamb-test - Test data from ILAMB (Needed by test suite)
 * ilamb - Land reference datasets (used internally to ilamb)
 * iomb - Ocean reference datasets (used internally to ilamb)
 * pmp-annual-cycle - Annual cycles post-processed from obs4MIPs data (To decide how it is used)

## Checklist

- [x] Tests added


packages/ref-core/src/cmip_ref_core/dataset_registry/__init__.py
```diff
@@ -14,7 +14,7 @@
 import pooch
 from loguru import logger
 
-DATA_VERSION = "v3.9"
+PMP_DATA_VERSION = "v3.9"
```
bouweandela, line 17: This could benefit from a more plugin-like architecture, where the cmip_ref_core package provides the infrastructure and the actual content is provided by the metrics packages.

lewisjared: This PR got a bit big with that change, but this adds a lot more consistency across the packages

lewisjared:
This now fetches the obs4REF data from the S3 bucket instead of PMP's servers

lewisjared:
@nocollier FYI This reworks and unifies the dataset fetching. I'm assuming that some of the datasets in your dataset registries should be moved to obs4ref registry? Then we can integrate the datasets into the sample data workflow and that should allow you to specify the required datasets via a `DataRequirement`

nocollier:
Awesome, I will take a look and will migrate the data I need.

lewisjared:
I'll need to upload the dataset to the bucket so let me know the datasets you need to migrate. We can do it in two steps, upload and add to the obs4ref catalog, then migrate to use the obs4mips data requirements

lee1043 (approved):
Looks good to me, thank you for this addition for the PMP's climatology reference datasets! The 20CR and HadISST monthly time series are in process to be published to the ESGF node but during the mean time, they are good to keep as part of this PR.

lewisjared:
@lee1043 I'm pretty keen to keep a single source of obs data for the May release. So even if these data are available via ESGF, I'd prefer if we had a single command to fetch them. For May that won't be obs4MIPs, but will be an archive on CEDA.

Please let me know if there are other obs files you need (hashes + URLs) and I will add them to the archive.

Don't stop getting those data on ESGF though, because we need to move away from this temporary fix!

PR #229, opened 2025-04-03
test: Add integration tests for pmp diagnostics

lewisjared:
## Description

Adds an integration test for PMP's extratropical modes of variability metric and adds the PMP metrics to the integration test suite.

The integration tests only run if the argument `--slow` is provided to pytest (either via `PYTEST_ADDOPTS` environment variable or the command line). 

This actually runs the metric (unlike the unit test). @lee1043  This might be useful for your debugging 

## Checklist

- [x] Tests added


lee1043 (approved):
This is great, thank you for adding this!

PR #232, opened 2025-04-06
Capture execution logs

lewisjared:
## Description

Redirects the logs of an execution to `out.log` for each execution. These logs will be displayed via `ref executions inspect` and via the web portal. 

Each metric provider has their own way of logging output, but these logs are intended as a super set of the provider logs (captured during a subprocess call) and the REF-specific logs.

## Checklist

- [x] Tests added


packages/ref/tests/unit/test_config.py
```diff
@@ -127,13 +127,15 @@ def test_defaults(self, monkeypatch):
         without_defaults = cfg.dump(defaults=False)
 
         assert without_defaults == {
+            "log_level": "INFO",
```
lewisjared, line 130: I have no idea why cattrs thinks this is a default value.

We don't use the "no-defaults" output so it isn't a major

PR #233, opened 2025-04-07
Use a shorter dataset key which is more stable

lewisjared:
## Description

Replaces the current implementation of a dataset key with a more stable alternative. The need for this key comes from determining which MetricExecutionGroup a collection of datasets is linked to. This key must be stable for this group as new data become available (new models, new versions, ancillary datasets, redactions)

Old: `dataset1_ACCESS-ESM1-5_tas_piControl_r1i1p1f1_dataset2_ACCESS-ESM1-5_areacella_piControl_r1i1p1f1`
New: `cmip6_piControl_ACCESS-ESM1-5_rsut_r1i1p1f1`

The biggest change is that the key is determined using the original groupby values, instead of after the constraints have been applied. The constraints are there to supplement the datasets in a group, but shouldn't cause the identity of the group to change. e.g. if the areacella wasn't present for the original execution it would still generate the same key, but on subsequent solves if it was available the same key would be used and it would replace the old results (we still keep the old version of results in the DB, but they would now no longer be shown unless you wanted a very specific result)

Applying constraints of a given group should not result in a different dataset key. If the intention is make a new group via constraints where the results don't replace the previous version of this group then that likely indicates that the groupby dimensions need to be modified.

You could come up with a set of constraints that generate the same datasets as another group. I'm not sure yet if that is a logic error that we need to guard against or is valid.

Grouping these executions at solve time and in the db layer in my opinion is a simpler solution than trying to resolve the "latest" executions after the fact. This approach enables the tracking of previous versions of an execution that replace each other.

## Checklist

- [ ] Tests added


mikapfl:
> Grouping these executions at solve time and in the db layer in my opinion is a simpler solution than trying to resolve the "latest" executions after the fact. This approach enables the tracking of previous versions of an execution that replace each other.

This is actually an interesting question. Since we are doing the grouping essentially for presentation purposes (not for purposes of determining what to run), why do you think that we should do this already at solve time? If we decouple this and build the execution groups after the fact, when presenting, we will be more agile. For example, we might end up wanting to change the grouping rules if we find that our grouping wasn't entirely perfect when we crunched all of CMIP6, but now we want to fix the grouping for showing the results. If we generally have a "dumb" db which only contains executions (with all associated metadata), we can re-build the groups easily. So, I'd like to know what advantages you see for building groups early.

packages/ref/src/cmip_ref/solver.py
```diff
@@ -41,61 +51,75 @@ class MetricExecution:
...
 
-    def build_metric_execution_info(self, output_root: pathlib.Path) -> MetricExecutionDefinition:
+    @property
+    def dataset_key(self) -> str:
         """
-        Build the metric execution info for the current metric execution
+        Key used to uniquely identify the execution group
+
+        This key is unique to an execution group and should be stable if new datasets are added
+        that
```
mikapfl, line 60: This sentence seems incomplete

lewisjared:
> Since we are doing the grouping essentially for presentation purposes (not for purposes of determining what to run),

That isn't strictly true.

`MetricExecutionGroup.should_run` checks if given a new dataset hash should be executed. One check is to determine if a group has been marked as dirty and therefore should be rerun even if it was run previously. We don't yet expose the ability to mark groups as dirty, but that feels like a natural way of marking parts of the problem domain as needing to be recalculated.

We could still achieve the same thing without the `MetricExecutionGroup`, but I would expect the implementation to be more complex than what is here.

We can add the selectors as JSON since we don't need to query against them at the moment. If we needed them for determining the groups at query-time we would need an additional table. Or maybe even one table per data source if we didn't just use a key-value table, but I would guess this makes it harder to write the SQL. 

> So, I'd like to know what advantages you see for building groups early.

Context and complexity. We would (and probably should anyway) need to add a representation of these selectors to the DB, to be able to construct the dataset name later. So we already have a dumb db.

I'm assuming that calculating groups at query-time seems like complex SQL that needs to happen on every query? I'm much less familiar with SQL than Python so an extra table seems like a more natural and maintainable solution. We would need to capture all the same information in the DB so the MetricExecutionGroup is effectively a cache of this information.

It doesn't block the ability to regenerate the groups at a later date, but it just seems to be a more complex approach than calculating groups at solve-time. If it proves to be wrong we can drop the group concept and recreate at query-time with relative ease.

I would prioritise tables that map one-to-one with our key concepts even if it isn't strictly normalised, over something more implicit.

lewisjared:
This change was replacing a temporarly solution in #65

mikapfl:
Generally, I see your points and definitely, the new implementation of dataset keys is better in the context of the current architecture.

> MetricExecutionGroup.should_run checks if given a new dataset hash should be executed.

I think this is pointing to some of my uneasiness with the current architecture. If I ingest 15 different datasets which all yield the same dataset key and the solve, only the latest one will be executed, right? Is that actually what we want?


> I'm assuming that calculating groups at query-time seems like complex SQL that needs to happen on every query?

No, we could also do this after solving and executing in python. So, a possible architecture is "solve executions" -> "execute them, put dumb results into DB" -> "group executions into MetricExecutionGroups, put groups into DB". This would at least decouple the groups from execution decisions, so that when we change the definition of groups, we don't have to dirty executions or whatever couplings we find.

mikapfl (approved):

lewisjared:
>  If I ingest 15 different datasets which all yield the same dataset key and then solve, only the latest one will be executed, right? Is that actually what we want?

How do they yield the same dataset key? Different versions? Then, yes only the latest would be run.

> No, we could also do this after solving and executing in python. So, a possible architecture is "solve executions" -> "execute them, put dumb results into DB" -> "group executions into MetricExecutionGroups, put groups into DB". This would at least decouple the groups from execution decisions, so that when we change the definition of groups, we don't have to dirty executions or whatever couplings we find.

Ahh I thought you were just referring to lazy group calculations. That is a possible alternative. I'm assuming you could operate on an individual group and not require a fan-in step that runs after all of the async executions?

There may be cases where just using the hash of the dataset's isn't sufficient and the initial groupings are useful information. A made-up example would be if you wished to calculate stats for a given variable with respect to some other supplementary datasets. You may end up with the following groups:
* tas with pr as a supplementary dataset
* pr with tas as a supplementary dataset

Those may end up with the same hash. Do we make that a metric provider problem to resolve or is it more reasonable to execute the metric twice because they would have different dataset keys as implemented here. The difference between those two executions could be determined with the selectors and/or the dataset key.

The data requirement might look something like:
```
DataRequirement(
    source_type="cmip6",
    filters=[FacetFilter(facets={"variable_id": ("tas", "pr")})],
    group_by=("variable_id", "source_id", "variant_label"),
    constraints=[
        AddSupplementaryDataset.from_defaults("tas", "cmip6"),
        AddSupplementaryDataset.from_defaults("pr", "cmip6"),
    ]
)
```

So if we do refactor, we shouldn't just rely on the dataset hashes, but then we are pretty much calculating groups before execution 🤷

lewisjared:
I'll add the scalars to the DB in a followup PR

PR #236, opened 2025-04-09
Add option to install a development version of a metrics package

bouweandela:
## Description

Add an option to install a development version of a metrics package. To add a development version of the metrics package used by a provider, instantiate the provider with a `repo` and `tag_or_commit` arguments, e.g.:
https://github.com/bouweandela/climate-ref/blob/38547119fad35a2079a9fd2653d02527d7ed45e5/packages/ref-metrics-esmvaltool/src/cmip_ref_metrics_esmvaltool/__init__.py#L11-L15

## Checklist

- [x] Tests added

### DIFF
--- a/changelog/236.feature.md
+++ b/changelog/236.feature.md
@@ -1,0 +1,1 @@
+Added the option to install development versions of metrics packages.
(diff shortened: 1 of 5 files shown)


bouweandela:
@lewisjared @mikapfl What do you think about this approach? Do we need to expose the functionality on the command line? Or would keeping it internal be better? That would keep the command line simpler and users can change the environments anyway after they've been created.

mikapfl:
I think it's relatively likely that users will want to use dev versions of metrics providers, at least in the next 12 months for metrics providers with a relatively slow release schedule. So, I like this approach.

mikapfl (approved):

PR #238, opened 2025-04-10
Obs doi

dhegedus99:
## Description
Adding columns within the database for global attributes related to complex citation and uncertainty information.

## Checklist

- [ ] Tests added


dhegedus99:
Test are still needed, but I wanted to create a draft PR so the REF is aware before I transition to part-time in my placement.

packages/ref/src/cmip_ref/datasets/obs4mips.py
```diff
@@ -145,6 +188,9 @@ class Obs4MIPsDatasetAdapter(DatasetAdapter):
         "units",
         "vertical_levels",
         "source_version_number",
+        "doi",
+        "references",
+        "tracking_id",
```
lewisjared, line 193: We should probably track the tracking id for CMIP6 datasets as well

lewisjared (commented):
There might be some regression files that need to be updated
I generally do this via
```
export PYTEST_ADDOPTS="--force-regen"
make test
```

PR #239, opened 2025-04-10
Add sea ice area seasonal cycle metric

bouweandela:
## Description

Add [sea ice area seasonal cycle metric](https://docs.esmvaltool.org/en/latest/recipes/recipe_ref.html).

## Checklist

- [ ] Tests added

### DIFF
--- a/packages/ref-metrics-esmvaltool/src/cmip_ref_metrics_esmvaltool/metrics/__init__.py
+++ b/packages/ref-metrics-esmvaltool/src/cmip_ref_metrics_esmvaltool/metrics/__init__.py
@@ -2,6 +2,7 @@
 
 from cmip_ref_metrics_esmvaltool.metrics.ecs import EquilibriumClimateSensitivity
 from cmip_ref_metrics_esmvaltool.metrics.example import GlobalMeanTimeseries
+from cmip_ref_metrics_esmvaltool.metrics.sea_ice_area_seasonal_cycle import SeaIceAreaSeasonalCycle
 from cmip_ref_metrics_esmvaltool.metrics.tcr import TransientClimateResponse
 from cmip_ref_metrics_esmvaltool.metrics.tcre import TransientClimateResponseEmissions
 from cmip_ref_metrics_esmvaltool.metrics.zec import ZeroEmissionCommitment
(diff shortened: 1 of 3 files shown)


lewisjared:
How do you see the handling of these non-obs4MIPs working for the production version? Do we have an archive of these that we can fetch (the ilamb + pmp approach), or do we have to rely on them being locally processed? Have you had any thoughts there

bouweandela:
If they will not be available from obs4MIPs in time for the AR7 deployment, we will need to prepare the archive and make it available somewhere. There is no website where they can be served from because of licensing restrictions. Maybe we can talk about this in some more detail at tomorrow's meeting.

lewisjared:
Yes lets discuss tomorrow. There will be a CEDA-hosted archive of obs data in time for the beta release. All the data that should be used in the REF will have been verified to have an open license that would permit this type of reuse.

bouweandela:
I have uploaded the required data and added the registry in this pull request. I did not see an easy way to find out the `--output-directory` specified in the `ref datasets fetch-data` command and use that from the `cmip_ref_metrics_esmvaltool` package to configure the path automatically. Would this need to be documented somewhere? Or should this be improved so the data can be found automatically?

lewisjared:
I think it is fine to document as an extra step for now. Having provider hooks might be a nice solution general solution to handling one-off tasks for a provider, but more work than just adding an extra line to the docs.

The other option is that you instatiate your registry and call fetch on the files you need. First time that will download the files, and then for the next execution it would return the filename of the cached result. That would only work if you didn't need the directory structure like you do with CMIP datasets

bouweandela:
I think I've found a good tradeoff now, by using the path to where the data has been fetched in the configuration automatically, but not automatically downloading. It looks like fetching causes a computation of a hash for all files in the cache, which would be too slow to do for every execution.

lewisjared (approved):
We will need some general documentation cleanup once everything had landed

bouweandela:
I suspect the integration test is failing because the pull request with sample data hasn't been merged yet: https://github.com/Climate-REF/ref-sample-data/pull/26

lewisjared:
I think the integration test failure is due to the reference data not being available

bouweandela:
It looks like you fixed that in #265, right?

lewisjared:
Yes. This is resolved now

Issue #234, opened 2025-04-07
Implicit statistic dimension


lewisjared:
## The problem
There is occasionally an implicit statistic dimension in some CMEC bundles, including PMP and some of the examples.
This is very annoying when we want to assert that a Metric bundle is valid.
At the moment, we don't check at all, and we only check the first n levels of the result follow the schema.  

Is there any reason for this? Not all models would have a "statistic", i.e. I wouldn't call ESMValTools Equilibrium Climate Sensitivity value a statistic. Why be implicit when we have the data structures to capture the all dimensions.

I think we need to strongly type and verify the dimensions and their values in the REF because, otherwise, it will turn into a mess very quickly. Enforcing this will make everyone's lives easier because we don't need to build around package-specific oddities. We should have a test suite that asserts the validity of these metric bundles and rejects erroneous metric bundles early in the development cycle.

The CVs are how we enforce required dimensions for a project

Related to #106 

## Definition of "done"
* [ ] Remove the option for implicit dimensions
* [ ] Add a test where more dimensions than specified are provided
* [ ] Update the test data in the repo to not have an implicit statistic dimension

lewisjared:
This is blocking a tidier implementation of #185

lewisjared:
@minxu74 Could you take a look at this some time? It's pretty easy to write metric bundles which pass the tests, but should fail if we were to be strict about all of the dimensions being registered.

minxu74:
@lewisjared Sorry. just see the issue from the email notifications. In early days, we included "statistics" in our results (i.e. in the `DIMENSIONS` and `json_structure`). But it generally is the deepest dimension, so it was considered as an unnecessary dimension as the keys of the deepest dictionary can represent their meanings. Moreover, metrics may have different `statistics`.  In previous, we did not enforce that a metric shall include all statistics defined in 'DIMENSIONS` even though the metrics did not have some of them. So, in conclusion, we could remove the implicit treatment of the deepest dimension, and enforce metrics to include all statistics (i.e. set the statistics that the metric do not have to a default value).  What do you think of it? Thanks.

lewisjared:
I think that all dimensions should be explicitly declared as it leads to less ambiguity. The innermost keys aren't any different from the other dimensions IMHO.

> enforce metrics to include all statistics (i.e. set the statistics that the metric do not have to a default value

Why do we need to enforce this?

minxu74:
[...]

PR #246, opened 2025-04-14
Remove implicit treatment of the deepest dimension (e.g., statistic)

minxu74:
## Description

In pycmec, we intentionally skip validations of the deepest dimension and omit the dimension in the `json_structure`, but the implicit treatment of the deepest dimension caused confusions and made the REF CV validation hard (see discussions in the issue #234 ). So this PR is to remove the implicit treatment and enforce its validations in pycmec.

## Checklist

- [x] Tests added

fixes: #234


minxu74:
@lee1043 I am trying to enforce a validation on the deepest dimension that we used not to check. The validation includes the test on the keys of the deepest dimension of the `RESULTS` are the same as the keys of the deepest dimension (`statistic` for PMP) defined in `DIMENSIONS`. PMP seemed to fail in this test. The `RESULTS` has the key `tcor_cbf_vs_eof_pc` that is not in the `statistic` dimension definition.  Also the `RESULTS` does not have the `frac_cbf_regird`.  Currently, we enforce the very strict rule of the identical keys of the deepest dimension in `RESULTS` and `DIMENSIONS`. We could loose the rule to allow the keys in `RESULTS` missing if needed. 

```
E       pydantic_core._pydantic_core.ValidationError: 1 validation error for CMECMetric
E         Value error, Dimension key mismatch in 'statistic' and level 6
E       Actual keys: ['bias', 'bias_glo', 'cor', 'cor_glo', 'frac', 'mean', 'mean_glo', 'rms', 'rms_glo', 'rmsc', 'rmsc_glo', 'stdv_pc', 'stdv_pc_ratio_to_obs', 'tcor_cbf_vs_eof_pc']
E       Expected keys: ['bias', 'bias_glo', 'cor', 'cor_glo', 'frac', 'frac_cbf_regrid', 'mean', 'mean_glo', 'rms', 'rms_glo', 'rmsc', 'rmsc_glo', 'stdv_pc', 'stdv_pc_ratio_to_obs']
E       Full actual structure:
E       [
E         "bias",
E         "bias_glo",
E         "cor",
E         "cor_glo",
E         "frac",
E         "mean",
E         "mean_glo",
E         "rms",
E         "rms_glo",
E         "rmsc",
E         "rmsc_glo",
E         "stdv_pc",
E         "stdv_pc_ratio_to_obs",
E         "tcor_cbf_vs_eof_pc"
E       ]
E       
E       Full expected structure:
E       {
E         "bias": {},
E         "bias_glo": {},
E         "cor": {},
E         "cor_glo": {},
E         "frac": {},
E         "frac_cbf_regrid": {},
E         "mean": {},
E         "mean_glo": {},
E         "rms": {},
E         "rms_glo": {},
E         "rmsc": {},
E         "rmsc_glo": {},
E         "stdv_pc": {},
E         "stdv_pc_ratio_to_obs": {}
E       } [type=value_error, input_value={'ACCESS-ESM1-5': {'r1i1p...get_model_eofs': 1}}}}}}, input_type=dict]
E           For further information visit https://errors.pydantic.dev/2.10/v/value_error
```

lewisjared:
At a minimum, we should be checking that the keys are declared in dimensions. I don't mind if we relax the requirement to include all values of a dimension for all dimensions as long as all dimensions are adequately described. That works fine with the flattened DB representation of the metrics.

#185 will try and process the metric values from a metric bundle and warn if it can't, but won't necessarily fail if something goes wrong when comparing against the controlled vocabulary.

packages/ref-core/src/cmip_ref_core/pycmec/metric.py
```diff
@@ -158,7 +159,17 @@ def _check_nested_dict_keys(cls, nested: dict[Any, Any], metdims: dict[Any, Any]
             dict_key.remove(MetricCV.ATTRIBUTES.value)
 
         if not (Counter(list(metdims[dim_name].keys())) == Counter(dict_key)):
-            raise ValueError("Error in dicts of Results")
+            # -raise ValueError("Error in dicts of Results\n"+
+            # -    json.dumps(metdims[dim_name].keys()) + "\n"
+            # -    json.dumps(dict_key)
+            # -)
```
lewisjared, line 165:
```suggestion
```

packages/ref-core/src/cmip_ref_core/pycmec/metric.py
```diff
@@ -167,10 +178,18 @@ def _check_nested_dict_keys(cls, nested: dict[Any, Any], metdims: dict[Any, Any]
...
-                tmp = dict(value)
-                if MetricCV.ATTRIBUTES.value in tmp:
-                    tmp.pop(MetricCV.ATTRIBUTES.value)
-                StrNumDict(tmp)
+                raise ValueError(
+                    f"Reached the last value {key} of the json_structure,\n"
+                    "but find there are some deeper dictionaries under it"
+                )
+            elif (not isinstance(value, float)) and (not isinstance(value, int)):
+                raise ValueError("Metric values have to be numeric types (float or int)")
```
lewisjared, line 186: FYI this is different to the type declared by `StrNumDict` (and preferable for the frontend because it is easier to format). Is that intentional?

i.e all leaf values must be floats or int

packages/ref-core/src/cmip_ref_core/pycmec/metric.py
```diff
@@ -167,10 +178,18 @@ def _check_nested_dict_keys(cls, nested: dict[Any, Any], metdims: dict[Any, Any]
...
+                # -tmp = dict(value)
+                # -if MetricCV.ATTRIBUTES.value in tmp:
+                # -    tmp.pop(MetricCV.ATTRIBUTES.value)
+                # -if any(isinstance(val, dict) for val in tmp.values()):
+                # -else:
+                # -    StrNumDict(tmp)
```
lewisjared, line 192:
```suggestion
```

packages/ref-core/src/cmip_ref_core/pycmec/metric.py
```diff
@@ -314,7 +333,10 @@ def _fill(cls, mdict: dict[Any, Any], mdims: dict[Any, Any], level: int = 0) ->
         dim_name = mdims[MetricCV.JSON_STRUCTURE.value][level]
         for key in mdims[dim_name].keys():
             if key not in mdict:
-                mdict[key] = {}
+                if level == len(mdims[MetricCV.JSON_STRUCTURE.value]) - 1:
+                    mdict[key] = -999.0
```
lewisjared, line 337: This should be hoisted up as a constant. I dislike these kinds of constants that exist in the same space as the values.

packages/ref-core/src/cmip_ref_core/pycmec/metric.py
```diff
@@ -386,8 +408,8 @@ def iter_results(self) -> Generator[MetricValue]:
         dimensions = cast(list[str], self.DIMENSIONS[MetricCV.JSON_STRUCTURE.value])
         # TODO: This is pretty hacky
         # A missing dimension in the results should be a validationError
-        if "statistic" not in dimensions:
-            dimensions = [*dimensions, "statistic"]
+        # -if "statistic" not in dimensions:
+        # -    dimensions = [*dimensions, "statistic"]
```
lewisjared, line 412:
```suggestion
```

minxu74:
@lee1043 yes. we could relax the rule and allow the missing keywords in the `RESULTS`.

packages/ref-core/src/cmip_ref_core/pycmec/metric.py
minxu74: If we allow the missing keywords in the deepest dimension, the if statements including the -999 statement can be removed.

packages/ref-core/src/cmip_ref_core/pycmec/metric.py
minxu74: Yes. We want to make sure the values in the deepest dictionary (i.e., the calculated metrics) are numeric values (i.e. float or int). Previously, we may not enforce the rule.

packages/ref-core/src/cmip_ref_core/pycmec/metric.py
minxu74: Thanks. I will cleanup code by removing these code comments.

lewisjared:
I think it should still be valid even if it is empty. That is an extreme case of missing keys

packages/ref-core/src/cmip_ref_core/pycmec/metric.py
lewisjared: Not needed anymore since we don't need to support missing values

packages/ref-core/tests/unit/test_metrics.py
```diff
@@ -70,23 +70,28 @@ def cmec_right_metric_dict():
...
+                "bias": {"name": "mean bias", "units": "inherit"},
+                "rmse": {"name": "root mean squre error", "units": "inherit"},
+            },
         },
         "RESULTS": {
             "E3SM": {
-                "Ecosystem and Carbon Cycle": {"overall score": 0.11, "bias": 0.56},
-                "Hydrology Cycle": {"overall score": 0.26, "bias": 0.70},
+                "Ecosystem and Carbon Cycle": {"overall score": 0.11, "bias": 0.56, "rmse": -999.0},
+                "Hydrology Cycle": {"overall score": 0.26, "bias": 0.70, "rmse": -999.0},
```
lewisjared, line 82: Perhaps these changes can be reverted as we don't have the concept of a missing value anymore?

lewisjared (approved):
Not sure why the test suite isn't running, but this needs a changelog, fixing the conflicts and cleaning up the example data then it is ready to go

minxu74:
There is a test fail in `ref` because the `CalendarMonths` is not consistent to the CMEC standard.  Currently, the pycmec only allow numeric for the values of the deepest dimension in the `RESULTS`. @lee1043 @lewisjared Shall we loose the restriction too?

I am going to merge the PR to the main. So I temporally loosed the restriction and let the test pass. Also there is a test fail in the PMP test due to the unknown dimension values in the `RESULTS` (i.e., the dimension values are not defined in the `DIMENSIONS`). We will address the problems in the following PRs. 

```
    "RESULTS": {
        "v3-LR": {
            "default": {
                "0101": {
                    "global": {
                        "bias_xy": {
                            "ann": 0.305,
                            "djf": 0.267,
                            "mam": 0.28,
                            "jja": 0.39,
                            "son": 0.282,
                            "CalendarMonths": [
                                "0.252",
                                "0.255",
                                "0.222",
                                "0.293",
                                "0.323",
                                "0.386",
                                "0.421",
                                "0.364",
                                "0.289",
                                "0.294",
                                "0.262",
                                "0.293"
                            ]
                        },

```

PR #248, opened 2025-04-14
Add selectors

lewisjared:
## Description

Adds a column to track group selectors.
This is useful for the front-end and if we need to recalculate dataset keys in future

## Checklist

- [ ] Tests added
- [ ] #185


packages/ref/src/cmip_ref/models/metric_execution.py
```diff
@@ -58,6 +58,15 @@ class MetricExecutionGroup(CreatedUpdatedMixin, Base):
     updated since the last execution.
     """
 
+    selectors: Mapped[dict[str, Any]] = mapped_column(default=dict)
+    """
+    Collection of selectors that define the group
+
+    These selectors are the unique key, value pairs that were selected during the initial groupby
+    operation.
+    These are also used to define dataset key.
```
mikapfl, line 67:
```suggestion
    These are also used to define the dataset key.
```

mikapfl (approved):

PR #250, opened 2025-04-15
Fix relative path in ESMValTool output bundle

lewisjared:
## Description

The integration tests where failing for two reasons,
* The path in the output bundle was relative to the inner result directory
* The copy file from scratch function didn't properly handle nested files

This also adds an integration test for running each metric individually and asserting that the results are correct. This test is marked as slow so requires adding `--slow` to the pytest arguments if you wish to run it (or manually run the [integration action for your branch](https://github.com/Climate-REF/climate-ref/actions/workflows/ci-integration.yaml) 

## Checklist

- [x] Tests added


packages/ref-metrics-esmvaltool/src/cmip_ref_metrics_esmvaltool/metrics/base.py
```diff
@@ -142,7 +142,7 @@ def build_metric_result(
             metadata = yaml.load(metadata_file.read_text(encoding="utf-8"))
             for filename in metadata:
                 caption = metadata[filename].get("caption", "")
-                relative_path = Path(filename).relative_to(result_dir.absolute())
+                relative_path = definition.as_relative_path(filename)
```
bouweandela, line 145: This didn't work for me when running the diagnostic with https://github.com/Climate-REF/climate-ref/blob/main/docs/how-to-guides/running-metrics-locally.py

lewisjared: I think I've resolved this in https://github.com/Climate-REF/climate-ref/pull/250/commits/ab0a27f293a58fdd96bea3e7d2aa015d951bc89d

There were some differing assumptions about relative/absolute paths

lewisjared: The creation of the output directory happens in the executors which is why it is needed here when running the metric directly

Issue #105, opened 2025-02-19
Repository migration


lewisjared:
## The problem
The cmip_ prefix in the repository name doesn't reflect the wider utility of this project beyond just CMIP. The community aspects of this project will be moved under the ESMO banner to reflect that.

I don't think this will cause too much impact for everyone as GitHub redirects urls, but cleaning up the cmip_ prefix will be very noisy so I want to do it when there are minimal PRs in flight. It is likely something I can get done in one day in my timezone to minimise the impact.

It would be useful to have this migration done before the hackathon to minimise the number of people impacted.

## Definition of "done"
* [x] Rename the repository (name Climate-REF)
* [x] Clean up any links #119 
* [x] Update docs to point to new projects
* [x] Final release
* [x] Refactor the `cmip_` prefix out of the package names.
* [x] Publish distributions under new name
* [x] Migrate documentation site to climate-ref. I think this is possible with redirects, but requires emailing the RTD admins

The proposed repository name is `climate-ref`, but any suggestions are welcome.

## Additional context
The CMIP AR7 FT related code can still live in the CMIP-REF organisation.

I'd prefer to move the package (the import names) to use `ref_` prefix. I'm less concerned about the distribution names (what is on pypi and conda) being different to the imports because we will likely have an AR7 FT REF package as the main entry point for installing packages. I think the repository name should match the distribution package prefixes though.

lewisjared:
The use of `metric` in the package names is also proving to be incorrect. I will drop this "nesting" during the migration i.e.

`cmip_ref_metrics_example` -> `ref_example` or `climate_ref_example`

lewisjared:
We should do this after a release to the old package names that redirect to the new packages

lewisjared:
Fixed as of v0.5.0

PR #269, opened 2025-05-01
Add deprecation notice in package README's

lewisjared:
## Description

Adds a deprecation notice to each Readme file on pypi with instructions about migrating. I will then do a final release using the old package names and then refactor to follow the new package names.

This no longer publishes the example package as it isn't needed now that we have actual providers.

## Checklist

- [ ] Tests added

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,6 @@ build: clean  ## build the packages to be deployed to PyPI
 	cp LICENCE NOTICE packages/ref
 	cp LICENCE NOTICE packages/ref-core
 	cp LICENCE NOTICE packages/ref-celery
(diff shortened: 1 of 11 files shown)


lewisjared:
One step of #105

lewisjared:
@nocollier @bouweandela @lee1043 Please scream if you have strong opinions about the naming.
The overview is that we replace cmip with climate in the package names and drop the `metrics` fragment. 

The "_" is also replaced with "-". I tried it using "_", but it felt a bit funny as a vast majority of python distributions use the "-" as a separator.

lewisjared:
I've renamed the packages in #270 according to this scheme

Issue #175, opened 2025-03-13
Failure when ingesting the whole of CMIP6 on JASMIN


mwklai:
## Describe the bug
When I tried to ingest the whole of CMIP6, it fails after ~20mins with the following error message. 

## Failing Test
Run the following command on JASMIN
`uv run ref -vv datasets ingest --source-type cmip6 /badc/cmip6/data/CMIP6/`

## Expected behavior
Expected all the metadata from the CMIP6 archive to be put into `.ref/db`.

## Screenshots
<details>
<summary>Full log output </summary>

```
[mwklai@sci-vm-04 climate-ref]$ uv run ref -vv datasets ingest --source-type cmip6 /badc/cmip6/data/CMIP6/
2025-03-13 14:28:56.528 | DEBUG    | cmip_ref.config:default:365 - Loading default configuration from /home/users/mwklai/workspace/climate-ref/climate-ref/.ref/ref.toml
2025-03-13 14:28:56.600 | INFO     | cmip_ref.database:__init__:79 - Connecting to database at sqlite:////home/users/mwklai/workspace/climate-ref/climate-ref/.ref/db/cmip_ref.db
2025-03-13 14:28:57.051 | DEBUG    | env_py:<module>:11 - Running alembic env
2025-03-13 14:28:57.066 | INFO     | alembic.runtime.migration:__init__:207 - Context impl SQLiteImpl.
2025-03-13 14:28:57.067 | INFO     | alembic.runtime.migration:__init__:210 - Will assume non-transactional DDL.
2025-03-13 14:28:57.333 | INFO     | cmip_ref.cli.datasets:ingest:98 - ingesting /badc/cmip6/data/CMIP6


╭──────────────────────────────────────────────────────── Traceback (most recent call last) ─────────────────────────────────────────────────────────╮
│ in pandas._libs.lib.maybe_convert_numeric:2391                                                                                                     │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
ValueError: Unable to parse string "nan"

During handling of the above exception, another exception occurred:

╭──────────────────────────────────────────────────────── Traceback (most recent call last) ─────────────────────────────────────────────────────────╮
│ /home/users/mwklai/workspace/climate-ref/climate-ref/packages/ref/src/cmip_ref/cli/datasets.py:112 in ingest                                       │
│                                                                                                                                                    │
[...]

PR #291, opened 2025-05-14
Support globbed directories

lewisjared:
## Description

Adds support for globbed directories when ingesting.
This is done by supporting multiple directories and relying on the globbing being done by the shell.

This won't solve the performance problem entirely, but it will definitely help. We might be able to parallelise across the input directories for a simple performance gain, but would need to benchmark that.

## Checklist

- [ ] Tests added


lewisjared:
Should help #175

PR #308, opened 2025-05-19
Add additional dimensions to the ESMValTool and ILAMB diagnostics

lewisjared:
## Description

Adds additional dimensions to the metric values.

For ESMValTool, `format_result` not returns the minimum set of dimensions that describe the metric values. The additional dimensions from the group_by definition are then added in `ESMValToolDiagnostic.build_execution_result`.

For ILAMB, we extract additional information out of the existing `source` then use that to generate the CMEC bundle.

## Checklist

- [ ] Tests added


lewisjared:
@bouweandela @nocollier FYI. 

The CMEC structure is cumbersome here. In both cases it would have been easier to stay with a flat representation of the values. 

The function to go from a list of values to a nested CMEC representation isn't that intuitive.

PR #310, opened 2025-05-19
Remove superfluous recipe keys in ECS diagnostic and rename diagnostic key

schlunma:
## Description

The ECS recipe produced by the REF contains lots of unused keys (in the original recipe, they have been used to define YAML anchors). This PR removes these keys.

In addition, it renames the diagnostic name from `cmip6/ecs` to `ecs/calculate`.

## Checklist

- [x] Tests added
- [ ] ~Documentation added (where applicable)~


lewisjared:
Thanks Manuel. Can you please add a changelog to in `/changelog`. There is a README.md file there to describe what is expected.

There is a conflict from #308, but that looks easy enough to resolve

lewisjared (approved):

lewisjared:
That is a failure in the regression test. I'll add the expected results. 

It involves running the integration test for the diagnostics as per [here](https://climate-ref.readthedocs.io/en/latest/development/#tests-and-code-quality) with the `--force-regen` argument

PR #325, opened 2025-05-23
Update ESMValTool version

### LINKED ISSUE #300: Comment on basic sea ice recipe
For the basic sea ice recipe (https://github.com/Climate-REF/climate-ref/blob/main/packages/climate-ref-esmvaltool/src/climate_ref_esmvaltool/diagnostics/sea_ice_area_seasonal_cycle.py) I would recommend to:

* rename the wrapper to something "sea-ice-area-basic" as "sea-ice-area-seasonal-cycle" does not cover everything the recipe is doing
* add a second timerange for the two "min" diagnostics (siarea_min): timerange = "1979/2014" to cover the whole observational time period available

bouweandela:
## Description

Update ESMValTool version

Closes #300

## Checklist

- [x] Tests added


bouweandela:
I have regenerated all the reference data: not all recipes/diagnostic scripts have changed, but figuring out what exactly has changed seemed more work than it's worth.

lewisjared (approved):

lewisjared:
> figuring out what exactly has changed seemed more work than it's worth

Yup having a time stamped folder in the regression output makes it a bit noiser than it needs to be.

PR #332, opened 2025-05-26
Fix ilamb burntfractionall

nocollier:
## Description

Adds a ILAMB diagnostic comparing model `burntFractionAll` to the GFED data product.

## Checklist

- [x] Tests added


nocollier:
@lewisjared I am not sure how tests are passing for this as I didn't think the reference data was in the S3 bucket.  I think you will need to put:

https://www.ilamb.org/ILAMB-Data/DATA/burntArea/GFED4.1S/burntArea.nc

to:

ilamb/burntFractionAll/GFED/burntArea.nc

Merge at will.

lewisjared:
It didn't fail because the CI for a PR doesn't run the slow tests (which take more like 30mins to run). So it passed because it was checking that the regression output could be correctly parsed, but it didn't try and run the diagnostic.

The daily integration tests check that works successfully. They can also be run manually for a given branch here: https://github.com/Climate-REF/climate-ref/actions/workflows/ci-integration.yaml

I'm merging a bunch of PRs now and then I'll run the integration tests to see what fails

lewisjared (approved):

PR #341, opened 2025-05-28
Enable retries hpcexecutor

minxu74:
## Description

To use the retries in parsl, the exception has to be raised to trigger it to rerun the failed tasks.  The PR changed the exception logic of `_process_run` and `join` methods in HPCExecutor.  The retries number can be set by the executor.config.retries in the `ref.toml`. The default value is 2. 

## Checklist

- [ ] Tests added

### DIFF
--- a/packages/climate-ref/src/climate_ref/slurm.py
+++ b/packages/climate-ref/src/climate_ref/slurm.py
@@ -1,4 +1,5 @@
 import importlib.util
+import os
 from typing import Any
 
 HAS_REAL_SLURM = importlib.util.find_spec("pyslurm") is not None
@@ -85,10 +86,13 @@ class SlurmChecker:
             return False
 
         sample_acc = account_info[0]
-        for acc in account_info:
-            if acc.user == "minxu":
-                sample_acc = acc
-                break
+        user_name = os.environ["USER"]
+
+        if user_name:
+            for acc in account_info:
+                if acc.user == user_name:
+                    sample_acc = acc
+                    break
 
         allowed_qoss = sample_acc.qos
         if allowed_qoss is None:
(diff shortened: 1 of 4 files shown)


packages/climate-ref/tests/unit/executor/test_hpc_executor.py
```diff
@@ -74,12 +74,28 @@ def test_join(self, metric_definition, tmp_path):
...
+        execution_result = ExecutionResult(
+            definition=metric_definition,
+            successful=False,
+            output_bundle_filename=None,
+            metric_bundle_filename=None,
+        )
+
+        future.set_exception(DiagnosticError("Some thing bad went wrong", execution_result))
+
+        executor.join(0.1)
```
lewisjared, line 91: Is there anything that you can assert about after joining?

lewisjared (approved):

packages/climate-ref/tests/unit/executor/test_hpc_executor.py
minxu74: Good point. I will add two assertions: (1) the result in the error is the same as the execution result, (2) the parsl_results shall be empty after join.

packages/climate-ref/src/climate_ref/slurm.py
```diff
@@ -85,10 +86,13 @@ def can_account_use_qos(self, account_name: str, qos_name: str) -> bool:
             return False
 
         sample_acc = account_info[0]
-        for acc in account_info:
-            if acc.user == "minxu":
-                sample_acc = acc
-                break
+        user_name = os.environ["USER"]
+
+        if user_name:
```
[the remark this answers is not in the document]
minxu74: Thanks.  I added a test to cover the block. There are other account_info that I directly used the first values w/o checking the `USER` environment variable and username in the `ref.toml`. I will submit another PR to fix them and document it. Currently, we have not used pyslurm to validate the executor configs yet as installing pyslurm need the path of slurm header and lib files that are different among HPCs. So pylsurm is not included in the REF virtual environment and the validation is off in default.

Issue #355, opened 2025-07-02
Broken integration tests


lewisjared:
## Describe the bug

#354 broke the integration tests. I think a manual step is needed in the integration test suite to install the providers

https://github.com/Climate-REF/climate-ref/actions/runs/16014357144/job/45178074020

PR #357, opened 2025-07-02
Install providers in integration tests

bouweandela:
## Description

Use the configured location for provider conda environments when running tests. This ensures conda environments are used from the expected location when running the tests.

The above also plays nicer with vscode because it doesn't create a lot of files in an unexpected place in the code repository (this can cause a slowdown and large memory consumption because vscode tries to index every file in the conda environment because the folder `.ref` is not in the default list of directories to ignore).

Pins scipy to < 1.16 (see https://github.com/statsmodels/statsmodels/issues/9584) to fix the import tests for the ilamb diagnostics package.

Closes https://github.com/Climate-REF/climate-ref/issues/355

## Checklist

- [ ] Tests added

### DIFF
--- a/.github/workflows/ci-integration.yaml
+++ b/.github/workflows/ci-integration.yaml
@@ -5,7 +5,7 @@ on:
   workflow_dispatch:
   # Run on each push to main and tagged version
   push:
-    branches: [main, install-providers-in-integration-tests]
+    branches: [main]
     tags: ['v*']
   # Runs every day at 2:15am (UTC) (~ midday in AEST)
   schedule:


bouweandela:
The conda environments are getting installed at
```
/home/runner/.config/climate_ref/software/conda/
```
but the tests expect them at
```
/home/runner/work/climate-ref/climate-ref/.ref/software/conda/
```

bouweandela:
@lewisjared Do you understand why the import tests have started failing? scipy is at https://github.com/Climate-REF/climate-ref/blob/7d82d396f0d4607a1f4775c2cdddbc2539aa12b6/uv.lock#L4047
on the `main` branch. Are we not running the tests with the environment from the lock file?

lewisjared:
Did the import tests fail here? In the log output for the integration test for main, I see `scipy==1.15.2` being installed which matches the lockfile.

lewisjared (approved):

bouweandela:
Yes, you can see it in the test of the first commit of this pull request, for example: https://github.com/Climate-REF/climate-ref/actions/runs/16024731592/job/45209793575

PR #358, opened 2025-07-03
Add support for NCI PBS Scheduler to the HPCExecutor

rbeucher:
## Description

This pull request introduces support for the job scheduler used on the Gadi supercomputer at NCI.

Gadi runs a customised version of PBSPro that differs from standard PBSPro installations by **disabling the `select` directive**, and instead requiring explicit use of flags like `-l mem=`, `-l ncpus=`, and `-l storage=`.

Because the existing `PBSProProvider` class in Parsl assumes support for `select`, it is incompatible with Gadi. To address this, a new provider class, `NCIGadiPBSProProvider`, has been added. This class handles Gadi's unique scheduler requirements and generates valid job submission scripts accordingly.

Note: I do not currently have access to a standard PBSPro system, so this implementation is specific to NCI Gadi. If needed, I am happy to collaborate on a broader/generalised solution that supports both flavours.

## Checklist

Please confirm that this pull request includes the following:

- [ ] **Tests added**: Existing `HPCExecutor` tests cover basic integration, but there are no scheduler-specific tests for SLURM or Gadi yet.
- [x] **Documentation added**, where applicable.
- [ ] **Changelog item added** in the `changelog/` directory.

### DIFF
--- a/docs/how-to-guides/hpc_executor.md
+++ b/docs/how-to-guides/hpc_executor.md
@@ -9,6 +9,7 @@ You could use HPCExecutor if:
   - The login nodes allow users to run a program for a long time like several hours with little computational resources (less than 25% of one CPU core and negligible memory)
   - You want to run REF under the HPC workflow i.e., submitting batch jobs
   - The scheduler on your HPC is __slurm__. We may include other schedules in the future if needed. Please make an [issue](https://github.com/Climate-REF/climate-ref/issues) describing your requirements.
+  - You are using NCI Gadi. Gadi uses a custom version of PBSPro. You can select the "nci" or "gadi" scheduler in the configuration file.
 
 ## Pre-requirements
 
@@ -38,6 +39,22 @@ walltime = "00:30:00"
 username = "minxu"
 qos = "debug"
 scheduler_options = "#SBATCH -C cpu"
+cores_per_worker = 1
+max_workers_per_node = 64
+```
+If you are using NCI Gadi, you can use the following configuration:
+```toml
+[executor.config]
+scheduler = "nci" # or "gadi"
+account = "m1234"
+queue = "normal"
+walltime = "00:30:00"
+username = "minxu"
+ncpus = 1
+mem = "4GB"
+jobfs = "10GB"
+storage = "/scratch/XX" # Make sure you have access to the storage
+scheduler_options = ""
 cores_per_worker = 1
 max_workers_per_node = 64
 ```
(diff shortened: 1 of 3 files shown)


rbeucher:
Hi @minxu74,

Thanks for reviewing.

I had the same thought and tried a few approaches using scheduler_options. The challenge is that the PBSProProvider class injects certain commands by default, and I found it quite difficult to work around that cleanly. I did consider overriding the template from the generic PBSProProvider, but that quickly made the code messy and harder to follow.

In the end, I felt that creating a dedicated class was the cleaner solution. From a user perspective, I also think it is more intuitive to specify the options explicitly in the ref.toml file, rather than dealing with a long string of flags that could easily lead to errors.

rbeucher:
In short, using the PBSProProvider means overriding the template as some options like `core, ncpus` `select` will remain.
I think that makes the code less readable, a clear separation is better in my opinion.

minxu74:
> I had the same thought and tried a few approaches using `scheduler_options`. The challenge is that the PBSProProvider class injects certain commands by default, and I found it quite difficult to work around that cleanly

Could you explain more about the challenge that you meet? Could you share the submit script under `runinfo` too when you use `scheduler_options`? Thanks.


> I also think it is more intuitive to specify the options explicitly in the ref.toml file

Yes. I agree. The `scheduler_options` is explicitly set in the `ref.toml`. The following is an example that I set for slurm provider on NERSC machine using "#SBATCH" that is similar to the "#PBS" in your code.

```
scheduler="slurm"
account="m2467"
req_nodes=1
walltime="00:30:00"
username="minxu"
qos="debug"
scheduler_options="#SBATCH -C cpu"
cores_per_worker=1
max_workers_per_node=64
```

minxu74:
> In short, using the PBSProProvider means overriding the template as some options like `core, ncpus` `select` will remain.
> I think that makes the code less readable, a clear separation is better in my opinion.

Thanks. I did not follow you here. using parsl PBSProProvider does not need to override the template as there is no template needed.

rbeucher:
`#PBS -l select=${nodes_per_block}:ncpus=${ncpus}${select_options}` 
Will be in the script and will cause issue in our case.

rbeucher:
Hi @minxu74,

I have generalised the approach and it should work for most PBS system.
Let me know what you think.

lewisjared (approved):
Awesome stuff @rbeucher

rbeucher:
Thanks both. I’d like to make this available to the ACCESS community at NCI, ideally without deploying from a fork. We have a workshop in a couple of weeks where I’d like to showcase it. Can we merge that PR? I can revisit the code afterwards based on your suggestions, @minxu74.

lewisjared:
I think we can merge this and then fix in a later release

PR #360, opened 2025-07-03
Use entrypoints to register provider plugins

bouweandela:
## Description

Use entrypoints to register provider plugins

This makes provider plugins findable by the REF if they have an entry like this in their `pyproject.toml` file:
```toml
[project.entry-points."climate-ref.providers"]
example = "climate_ref_example:provider"
```

## Checklist

- [x] Tests added


bouweandela:
> I tried removing unique=True from but that did not help.

It seems to have worked after all, only the smoke test keeps timing out now.

lewisjared:
I thought I had fixed that too, but must have missed the actual unique constraint.

Take a look at the developer docs: https://climate-ref.readthedocs.io/en/latest/development/#adding-a-database-migration

and let me know if there is any additional information we need to add there.

bouweandela:
I followed those instructions, very helpful, and that resulted in cb0742f455a471a4f2766043466a581539f38012. Does that look OK? If yes, then I think the instructions are fine as they are.

lewisjared:
That looks right to me. Not sure why the smoke test fails though. I've rerun it to see if it is just flakey

lewisjared:
I believe the smoke test is also trying to queue up the example provider as it is installed. The example provider doesn't have a docker container so hangs. That is almost by design as consumers may come and go, but the queue should remain

lewisjared:
@bouweandela I'll rework this in a follow up PR to replace the "package" argument with a "provider" argument. This can be used to refer to the entry points rather than indirectly using the package name. It's worth a breaking change

lewisjared:
Merging and then fixing the smoke test in #367

PR #362, opened 2025-07-07
chore: remove parsl from windows

lewisjared:
## Description

Removes installing Parsl on Windows and updated the readme to clarify that Windows isn't supported in general.
I opted to keep the Windows import tests in as while we don't support Windows, 
some brave souls may wish to try.
So our testing for Windows is limited to making sure it *might* be installable.

## Checklist

- [ ] Tests added


minxu74 (approved):
Cool. I have never thought of running REF on the  Windows OS as I have not used it for work since 2002. It is good to indicate that it is not officially supported while users can still try REF and parsl on windows machines.

PR #363, opened 2025-07-09
Demo use of series output

bouweandela:
## Description

Add an example of recording series output.

## Checklist

- [ ] Tests added

### DIFF
--- a/packages/climate-ref-core/src/climate_ref_core/metric_values/typing.py
+++ b/packages/climate-ref-core/src/climate_ref_core/metric_values/typing.py
@@ -6,6 +6,27 @@ from pydantic import BaseModel, model_validator
 Value = float | int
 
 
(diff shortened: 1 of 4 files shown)


packages/climate-ref-core/src/climate_ref_core/metric_values/typing.py
```diff
@@ -6,6 +6,27 @@
 Value = float | int
 
 
+class SeriesDefinition(BaseModel):
+    """
+    A definition of a 1-d array with an associated index and additional dimensions.
+    """
+
+    file_pattern: str
```
bouweandela, line 14: Maybe this is not generic enough to go here, as it assumes that output is stored in a file

lewisjared:
Closing in preference for #374 which shares these commits

PR #364, opened 2025-07-16
Update download URL for datasets

lewisjared:
## Description

Updates the URL for fetching the test and obs data.
This may resolve some of the odd SSL issues we have had in the past (I believe older versions of TLS are now supported) 
## Checklist

- [ ] Tests added

### DIFF
--- a/packages/climate-ref-core/src/climate_ref_core/dataset_registry.py
+++ b/packages/climate-ref-core/src/climate_ref_core/dataset_registry.py
@@ -18,7 +18,7 @@ from rich.progress import track
 
 from climate_ref_core.env import env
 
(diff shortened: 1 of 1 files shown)


lewisjared:
@bouweandela @minxu74 I think this might resolve some of the SSL issues. @minxu74 Could you try fetching the obs data on a node where you had SSL errors previously without your workarounds?

minxu74:
@lewisjared I just tested it and it can download the obs data without the SSL error. But when I ran 
```
from urllib.request import urlopen;print(urlopen('https://www.howsmyssl.com/a/check').read())
```
it is still failed.

lewisjared:
I think that is an improvement? That suggests that there are still SSL issues, but not with the ref??

minxu74:
It only happened in the REF uv venv, no problem with the conda venv and the python provided by the HPC.

PR #371, opened 2025-07-31
docs revisions

nocollier:
## Checklist

- [ ] Tests added


nocollier:
I have integrated all the comments that I can from the MB TT. In the end it was hard because some of what they said came from a poor concept of how REF does what REF does (our fault) or their desires to see some things work differently which is a bit out of scope. I have struck through the comments I addressed and left others for the time being.

I added 2 pages that I am not sure you will like, but I felt a lot of the comments came from common misconceptions in our community and so I tried to explain them:

- `nutshell.md` where I go into what REF does and the terms we use to talk about it. References to this file have been made in several places if users are confused. I suspect even I don't have it totally right so please read that carefully.
- `packaging_confusion.md` where I get into why there are so many package managers and why you need to know if you code is pure python or not. Just my POV, but I think it would help make the rest of the docs easier to understand. Please read this carefully too and make sure you agree.

I noticed that the jupyter cells here aren't executing and might be contributing to the docs not being understood:

https://climate-ref.readthedocs.io/en/latest/how-to-guides/dataset-selection/

Once I figure out why I can better address the comments on that page specfically.

lewisjared:
@nocollier I'm going to merge this now as I need to make a new release

PR #375, opened 2025-08-17
validate hpc slurm config using pydantic

minxu74:
## Checklist

- [x] Tests added

### DIFF
--- a/packages/climate-ref/src/climate_ref/executor/hpc.py
+++ b/packages/climate-ref/src/climate_ref/executor/hpc.py
@@ -12,9 +12,9 @@ try:
 except ImportError:  # pragma: no cover
     raise ImportError("The HPCExecutor requires the `parsl` package")
 
-import os
+import re
 import time
-from typing import Any
+from typing import Annotated, Any, Literal
 
 import parsl
 from loguru import logger
@@ -23,6 +23,7 @@ from parsl.config import Config as ParslConfig
 from parsl.executors import HighThroughputExecutor
 from parsl.launchers import SrunLauncher
 from parsl.providers import SlurmProvider
+from pydantic import BaseModel, Field, StrictBool, field_validator, model_validator
 from tqdm import tqdm
 
 from climate_ref.config import Config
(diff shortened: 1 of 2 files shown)


packages/climate-ref/src/climate_ref/executor/hpc.py
```diff
@@ -36,6 +37,78 @@
...
+            r"(?:\s+-\w+\s+[^\s]+)*)"  # Additional options
+            r"\s*$",
+            re.IGNORECASE | re.MULTILINE,
+        )
+
+        print(v)
+        for line in v.split("\n"):
+            print(line.strip().upper().startswith("#SBATCH"))
+
+            print(line, sbatch_pattern.match(line.strip()))
```
lewisjared, line 86: Are you meant to be printing this? Perhaps you format it as a string and write a debug log message?

packages/climate-ref/src/climate_ref/executor/hpc.py
```diff
@@ -36,6 +37,78 @@
 from .local import ExecutionFuture, process_result
 
 
+class SlurmConfig(BaseModel):
+    """Slurm Config Pydnatic model"""
```
lewisjared, line 41:
```suggestion
    """Slurm Config Pydantic model"""
```

This Pydantic-part is an implementation detail so I'm not sure if its needed

packages/climate-ref/src/climate_ref/executor/hpc.py
```diff
@@ -174,39 +244,41 @@ def _validate_slurm_params(self) -> None:
...
             raise ValueError(
-                f"Walltime: {self.walltime} exceed the maximum time "
-                f"{max_walltime_minutes} allowed by {self.partition} and {self.qos}"
+                f"Walltime: {self.slurm_config.walltime} exceed the maximum time "
+                f"{max_walltime_minutes} allowed by {self.slurm_config.partition} and {self.slurm_config.qos}"
             )
 
     def _initialize_parsl(self) -> None:
-        executor_config = self.config.executor.config
+        # executor_config = self.config.executor.config
```
lewisjared, line 252:
```suggestion
```

lewisjared (approved):
Some minor changes, but nice stuff. Should we do something similar for PBS?

packages/climate-ref/src/climate_ref/executor/hpc.py
minxu74: Yes. I will remove it.

packages/climate-ref/src/climate_ref/executor/hpc.py
minxu74: I will remove it.  Thanks

PR #408, opened 2025-09-17
Add groupby to require facets constraint

bouweandela:
## Checklist

- [x] Tests added


packages/climate-ref-core/src/climate_ref_core/constraints.py
```diff
@@ -152,6 +152,12 @@ class RequireFacets:
     required_facets: tuple[str, ...]
     operator: Literal["all", "any"] = "all"
 
+    group_by: tuple[str, ...] = ()
```
lewisjared, line 155: Perhaps the default should be None instead of an empty tuple to indicate that no grouping is happening? That would be similar to `DataRequirement.group_by`.

packages/climate-ref-core/src/climate_ref_core/constraints.py
```diff
@@ -152,6 +152,12 @@ class RequireFacets:
...
+    """
+    The fields to group the datasets by. Each group must contain all or any of the
+    required facets to fulfill the constraint.
```
lewisjared, line 158:
```suggestion
    required facets to fulfill the constraint.
    
    The default is to treat the datasets as a single group.
```

lewisjared (commented):
It might need an additional sentence or two in the docstring of the class describing this use case as I didn't understand it until I saw how you used it.

I don't think it's worth splitting out into a separate constraint as it is so similar.

lewisjared (approved):

PR #409, opened 2025-09-18
Make it possible to use & and | to combine facet filters

bouweandela:
## Description

Make it possible to use `&` and `|` to combine facet filters as recommended by @lewisjared in https://github.com/Climate-REF/climate-ref/pull/407#discussion_r2357325512

## Checklist

- [ ] Tests added


bouweandela:
@lewisjared Do we want to keep https://github.com/Climate-REF/climate-ref/blob/9dbb6a307642e1b375e009193f80490cf3e8a692/packages/climate-ref-core/src/climate_ref_core/diagnostics.py#L319-L325
as a tuple of filters or replace it with a single filter?

lewisjared:
If we keep the existing filters and move to `or` instead of `and` we end up with a broader set of initial datasets. That isn't a bad thing and can always be reduced in a constraint if need be.

I don't mind if we use `or` instead. The explicit inclusion is likely easier to understand than  the inverse so `keep=False` can be removed.

PR #412, opened 2025-09-22
Support adding/updating datasets

### LINKED ISSUE #316: Allow registering new files for a dataset in the database
## The motivation

When additional files become available for a dataset, it would be nice if those could be ingested in the database.

## The proposed solution

The command `ref datasets ingest` should check if a file is already available in the database and if not, add it, even when the dataset that the file is part of already exists in the database.

Related code: https://github.com/Climate-REF/climate-ref/blob/8c26336e4cc6cfa44a165d141602f6ebd5b01494/packages/climate-ref/src/climate_ref/datasets/base.py#L141-L144

lewisjared:
## Description

Reworks the ingest process to allow adding/updating existing datasets as metadata changes/new files are available.
This info is also aggregated and logged

Adding new files will cause the hash of any applicable `DatasetCollection`'s to change on the next solve. This would then cause a new execution to be performed.

## Checklist

- [x] Tests added

Closes #316


bouweandela:
We can remove `ERA-5` `ta` from obs4REF as it's available through obs4MIPs / ESGF, so undo #334. The reason I added it back then was that the ESGF server holding the data was offline all the time, but that has hopefully improved by now.

lewisjared:
I've tested this when I've been running the latest set of results and had no issues.

PR #420, opened 2025-09-26
Upgrade to include obs4MIPs fetching

lewisjared:
## Description

Updates the fetch script to include obs4MIPs data and ohc/thetaO which was missing in the initial solve.

## Checklist

- [ ] Tests added


scripts/fetch-esgf.py
```diff
@@ -0,0 +1,360 @@
...
+
+Request = CMIP6Request | Obs4MIPsRequest
+
+requests: list[Request] = [
+    CMIP6Request(
+        id="esmvaltool-climate-at-global-warmings-levels",
+        facets=dict(
+            variable_id=["pr", "tas"],
+            experiment_id=["ssp126", "ssp245", "ssp370", "ssp585", "historical"],
+        ),
```
bouweandela, line 105: You could reduce the data volume by specifying the `table_id` according to how the ESMValTool diagnostics do it, e.g. https://github.com/Climate-REF/climate-ref/blob/27b28bffd4d8ba13b294f3f0d003102f348a9e44/packages/climate-ref-esmvaltool/src/climate_ref_esmvaltool/diagnostics/cloud_radiative_effects.py#L41 and refined in #414

lewisjared: Added the extra table_id's. 

It would have been cool to explicitly link the data requirements, but doing it manually was likely quicker at this point in time

scripts/fetch-esgf.py
```diff
@@ -174,6 +180,7 @@ def fetch(self, remove_ensembles: bool = True):
                 "vegFrac",
             ],
             experiment_id=["historical"],
+            table_id=("Amon", "Emon", "Lmon"),
```
bouweandela, line 183: This will probably still fetch a bit too much data, see https://github.com/Climate-REF/climate-ref/pull/414, but maybe that's actually nice because it allows testing in a more real world scenario

lewisjared: I'm not too worried about making it super tight as long as it isn't ocean/atmos datasets. It doesn't make much of a difference if some of the more esoteric runs are also captured.

Issue #425, opened 2025-09-26
Investigate missing ESMValTool sea ice sensitivity


lewisjared:

lewisjared:
#416 resolved the dataset selection issue. There is now an execution, but it failed!

[out.log](https://github.com/user-attachments/files/22568813/out.log)
[execution-3361.log](https://github.com/user-attachments/files/22568831/execution-3361.log)

Looks like invalid grid info for `CAS-ESM2-0` during a `cmor_check_metadata` preprocessor.

We don't have any information in our tracked data that could be used to filter this out? Can you filter this out via ESMValTool during the recipe? Maybe it is time to come up with a simple approach for blacklisting some datasets

bouweandela:
Could you share the recipe, so I can try running it and find all the failing datasets? I added some log messages in #432 to make debugging in the future easier.

I'll see if I can add a feature to exclude some datasets, as ESMValTool is not able to do that. In the usual use case we count on recipe authors to take care of that. Of course, we could add it, but it would need some work.

PR #447, opened 2025-10-03
Add ignore datasets constraint and configuration file

bouweandela:
## Description

Add a constraint and configuration file that can be used to ignore specific datasets.

This is needed to support filtering out datasets that crash multi-model analyses as reported in #425.

## Checklist

- [x] Tests added


default_ignore_datasets.yaml
```diff
@@ -0,0 +1,10 @@
+esmvaltool:
```
bouweandela, line 1: @lewisjared Is this the approach you were looking for? Or would you prefer to have separate files per provider?

PR #454, opened 2025-10-06
Convert defaulf fillvalues to NaNs

bouweandela:
## Checklist

- [ ] Tests added


lewisjared:
NaNs aren't valid json. Can we ensure we write null instead in the json output.

lewisjared:
If you open up Chrome DevTools and run the following JS snippet `JSON.parse('{"a": NaN}')` in the console you get:

```
> JSON.parse('{"a": NaN}')
VM921:1 Uncaught SyntaxError: Unexpected token 'N', "{"a": NaN}" is not valid JSON
    at JSON.parse (<anonymous>)
    at <anonymous>:1:6
> JSON.parse("NaN")
VM1181:1 Uncaught SyntaxError: "NaN" is not valid JSON
    at JSON.parse (<anonymous>)
    at <anonymous>:1:6
```

and JSON stands for JavaScript Object Notation!

The fact that python doesn't coerce the NaN to null to be standard-following is quite annoying. I guess that null is closer to `None` and it's impossible to know if a value is meant to be a NaN or a `None`.

Hmm I agree that isn't the best error message. My guess is that it can't coerce a None to SeriesMetricValue which is `float | int`. I'm having a look now if there is any Pydantic options

lewisjared:
What if you change to `json.dump([s.model_dump(mode="json") for s in series], f, indent=2)`? mode defaults to "python".

lewisjared:
@bouweandela Maybe something like: https://github.com/Climate-REF/climate-ref/commit/90b1d8134327224f864d0bf07b15d704db734066

That stablely ordered the outputs

Issue #461, opened 2025-10-12
Failure due to the timestamp in the intermeidate results of the PMP annual cycle diagnostic


minxu74:
## The problem

Found that there was a timestamp attached to the intermediate results of the annual cycle diagnostic. For example:
```
/pscratch/sd/m/minxu/mytest_ref_2025-09-25/config/software/conda/micromamba run --prefix /pscratch/sd/m/minxu/mytest_ref_2025-09-25/config/software/conda/pmp-a0cd98ec96f7c858f788439a4f0b4951fe2d7a55 mean_climate_driver.py -p /pscratch/sd/m/minxu/mytest_ref_2025-09-25/climate-ref/packages/climate-ref-pmp/src/climate_ref_pmp/params/pmp_param_annualcycle_2-metrics.py --vars ts --custom_observations /pscratch/sd/m/minxu/mytest_ref_2025-09-25/config/scratch/pmp/annual-cycle/29d06681d9b0fe74e07ed58444b9fe23a1b20775/obs_dict.json --test_data_path /pscratch/sd/m/minxu/mytest_ref_2025-09-25/config/scratch/pmp/annual-cycle/29d06681d9b0fe74e07ed58444b9fe23a1b20775 --test_data_set GISS-E2-1-G --realization r4i1p1f1 --filename_template "%(variable)_GISS-E2-1-G_hist-GHG_r4i1p1f1_clims.198101-200512.AC.v20251011.nc" --metrics_output_path /pscratch/sd/m/minxu/mytest_ref_2025-09-25/config/scratch/pmp/annual-cycle/29d06681d9b0fe74e07ed58444b9fe23a1b20775
```

The intermediate file `%(variable)_GISS-E2-1-G_hist-GHG_r4i1p1f1_clims.198101-200512.AC.v20251011.nc` had the timestamp `v20251011`. So the above command was generated on Oct. 11, but the intermediate file maybe was generated on Oct. 12 with the timestamp `v20251012`.  

It will cause the execution failure and maybe the `IndexError`.

PR #465, opened 2025-10-14
PMP annual cycle: make flexible timestamp

lee1043:
## Description

Address #461 

## Checklist

- [ ] Tests added

### DIFF
--- a/changelog/539.fix.md
+++ b/changelog/539.fix.md
@@ -1,0 +1,1 @@
+Fixed confusion between variable_id and out_name in fake CMIP7 data.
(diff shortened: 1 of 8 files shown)


lewisjared:
I don't think this will fix the issue. A date code needs to be generated at the start of the run and that is used for all results consistently

minxu74:
@lee1043 The modification that you made was applied to the files with the `mon` in their filenames. But the error was caused by the file with the `clim` in its filename. Also the error was only triggered when the metric was started before the midnight, and the `clim` file was created after midnight.

lee1043:
Hi @minxu74, thank you so much for helping me on this PR. 

Could you help me finding the comment you mentioned in [your reply](https://github.com/Climate-REF/climate-ref/pull/465#issuecomment-3929432930)? I guess you meant you have added new comments to the code in addition to the [previous reply](https://github.com/Climate-REF/climate-ref/pull/465#issuecomment-3912011446), bud I was not able to find it on the repository. Did you add your comment to the code in the PR directly?

PR #497, opened 2026-01-28
test: add CI job to verify solve works without network access

lewisjared:
## Summary

- Add integration tests that verify solve operations work without network

This is the first step toward implementing provider lifecycle hooks. The tests are expected to fail (hence `continue-on-error: true`) until the hooks are implemented.


Makefile
```diff
@@ -130,6 +131,7 @@ test-quick: clean  ## run all the tests at once
 	# It doesn't execute each test using the target package as above
 	uv run \
 		pytest tests packages \
+		--ignore=packages/climate-ref/tests/integration/test_offline_solve.py \
```
lewisjared:
```suggestion
```

PR #498, opened 2026-01-28
feat(providers): add lifecycle hooks for offline execution setup

lewisjared:
## Description

Add provider lifecycle hooks that enable providers to prepare for offline execution on HPC compute nodes where internet access is unavailable during solve operations.

### New Lifecycle Hooks

Added to `DiagnosticProvider` base class:
- `setup()` - orchestrates all hooks in correct order
- `setup_environment()` - environment setup (e.g., conda)
- `fetch_data()` - pre-fetch reference data to pooch cache
- `post_setup()` - post-setup tasks
- `validate_setup()` - verify setup is complete

All hooks are **optional** (default no-op) and **idempotent** (safe to run multiple times).

### Provider Implementations

- **ESMValToolProvider**: fetches "esmvaltool" registry data
- **PMPDiagnosticProvider**: fetches "pmp-climatology" registry data
- **ILAMBProvider**: fetches "ilamb-test", "ilamb", "iomb" registry data

### CLI Command

New `ref providers setup` command with options:
- `--provider`: filter to single provider
- `--skip-env`: skip environment setup
- `--skip-data`: skip data fetching
- `--validate-only`: only validate setup status

### Usage

```bash
# On login node with internet access:
ref providers setup

# Then on compute node without internet:
ref solve
```

### Additional Changes

- Updated CI offline-solve workflow to use `providers setup`
- Fixed bug where network errors during ignore-datasets file fetch weren't caught

## Checklist

- [x] Tests added

### DIFF
--- a/packages/climate-ref-core/src/climate_ref_core/dataset_registry.py
+++ b/packages/climate-ref-core/src/climate_ref_core/dataset_registry.py
@@ -97,7 +97,7 @@ def validate_registry_cache(
             continue
 
         # Check if file exists in cache
-        cached_path = pathlib.Path(registry.abspath(key))  # type: ignore[attr-defined]
+        cached_path = registry.abspath / key  # type: ignore[attr-defined]
         if not cached_path.exists():
             errors.append(f"{name}: File not cached: {key}")
             continue
--- a/packages/climate-ref-core/tests/unit/test_dataset_registry/test_dataset_registry.py
+++ b/packages/climate-ref-core/tests/unit/test_dataset_registry/test_dataset_registry.py
@@ -8,6 +8,7 @@ from climate_ref_core.dataset_registry import (
     _verify_hash_matches,
     dataset_registry_manager,
     fetch_all_files,
+    validate_registry_cache,
 )
 
 NUM_OBS4REF_FILES = 58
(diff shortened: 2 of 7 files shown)


lewisjared:
## Follow-up work

After this PR merges, we should address the following in separate PRs:

1. **Move PMP climatology fetching and ingestion into PMP provider setup**
   - Currently users must run `ref datasets fetch-data --registry pmp-climatology` and `ref datasets ingest --source-type pmp-climatology` separately
   - The PMP provider's `fetch_data()` hook already downloads to pooch cache, but the data also needs to be ingested into the catalog for the solver
   - Ideally `ref providers setup` would handle both fetching AND ingesting for pmp-climatology

2. **Fetch cartopy data for ILAMB provider**
   - ILAMB uses cartopy for map plotting which requires shapefiles
   - The cartopy data download should be handled by the ILAMB provider's `fetch_data()` hook or `post_setup()` hook
   - Currently this is done via a separate script (`scripts/download-cartopy-data.py`)

3. **Fetch ESMValTool recipes**
   - ESMValTool recipes should be fetched during provider setup
   - This ensures all recipe files are available for offline execution

packages/climate-ref/src/climate_ref/cli/providers.py
```diff
@@ -82,3 +82,91 @@ def create_env(
...
+                provider_.setup_environment(config)
+            if not skip_data:
+                logger.info(f"Fetching data for {provider_.slug}")
+                provider_.fetch_data(config)
+            provider_.post_setup(config)
+            logger.info(f"Finished setting up provider {provider_.slug}")
+        except Exception as e:
+            logger.error(f"Failed to setup provider {provider_.slug}: {e}")
+            failed_providers.append(provider_.slug)
+
```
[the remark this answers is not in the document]
lewisjared: Fixed in commit 5dcead96 - Added validate_setup() call after setup completes, failing the provider if validation returns False

packages/climate-ref/src/climate_ref/cli/providers.py
```diff
@@ -82,3 +82,90 @@ def create_env(
...
+    This command prepares all providers for offline execution by:
+    1. Creating conda environments (if applicable)
+    2. Fetching required reference datasets to pooch cache
+
+    All operations are idempotent and safe to run multiple times.
+    Run this on a login node with internet access before solving on compute nodes.
+
+    Examples
+    --------
```
bouweandela: This doesn't render nicely:
```
...
 This command prepares all providers for offline execution by: 1. Creating conda environments (if applicable) 2. Fetching required reference datasets to pooch cache                                                                                          
 All operations are idempotent and safe to run multiple times. Run this on a login node with internet access before solving on compute nodes.                                                                                                                 
 Examples --------     # Setup all providers     ref providers setup                                                                                                                                                                                          
 # Setup only ESMValTool provider     ref providers setup --provider esmvaltool                                                                                                                                                                               
 # Only validate, don't run setup     ref providers setup --validate-only                                                                                                                                                                                     
...
```

packages/climate-ref/src/climate_ref/cli/providers.py
```diff
@@ -82,3 +82,90 @@ def create_env(
             logger.info(f"Skipping creating {txt} because it does not use conda environments.")
 
     list_(ctx)
+
+
+@app.command()
+def setup(
```
bouweandela: How about deprecating the `create-env` command, since it will no longer be needed and may cause confusion?

It be nice if `ref providers list` also shows the paths where data is installed.

PR #519, opened 2026-02-09
feat(esmvaltool): add CMIP7 support to all ESMValTool diagnostics

lewisjared:
## Summary

- Add CMIP7 as an alternative data source for all 16 ESMValTool diagnostics in `data_requirements`
- Add CMIP7 to recipe infrastructure for CMIP7 replacing the deprecated `drs` attribute
- Add `get_cmip_source_type()` helper so `update_recipe` methods dynamically select CMIP6 or CMIP7 input files

### Per-diagnostic pattern

Each CMIP7 `DataRequirement` mirrors its CMIP6 counterpart with:
- `frequency` filter instead of `table_id`
- `variant_label` instead of `member_id` in `group_by`
- `SourceDatasetType.CMIP7` for `AddSupplementaryDataset`


bouweandela:
We may also want to consider/check which the diagnostic authors which `branding_suffix` should be used for the variables in the data requirements. All of them? Only one? Different per variable?

lewisjared:
## CMIP7 ESMValTool Diagnostic Status

Tested all 18 ESMValTool diagnostics against CMIP7 data. **9 pass, 8 fail, 1 disabled.**

### Passing (9)

- cloud-scatterplots-clwvi-pr
- enso-basic-climatology
- global-mean-timeseries
- regional-historical-annual-cycle
- regional-historical-timeseries
- regional-historical-trend
- transient-climate-response
- transient-climate-response-emissions
- zero-emission-commitment

### Failing — blocked on ESMValCore upstream issues (8)

**CMIP7 CMOR tables missing (4 diagnostics):**
Variables like `lwcre`, `swcre`, and `rtnt` can't be resolved when `project='CMIP7'` because the mip value doesn't map to a loadable CMOR table. https://github.com/ESMValGroup/ESMValCore/issues/2980.
- climate-drivers-for-fire (`tasmax` / `atmos`)
- cloud-radiative-effects (`lwcre` / `atmos`)
- cloud-scatterplots-clivi-lwcre (`lwcre` / `atmos`)
- cloud-scatterplots-clt-swcre (`swcre` / `atmos`)
- equilibrium-climate-sensitivity (`rtnt` / `atmos`)

The other set of issues which are yet to be resolved is due to the fixes not being applied for these CMIP7 datasets. @bouweandela Is there a gold standard source that we could use that doesn't require fixes?

**CoordinateMultiDimError on curvilinear ocean grids (3 diagnostics):**
`area_statistics` fails on multi-dimensional coordinates (e.g., CanESM5 tripolar ocean grid `gn`) because `try_adding_calculated_cell_area` runs before using the provided `areacello`.
- enso-characteristics (CanESM5 `tos`)
- sea-ice-area-basic (CanESM5 `siconc`)
- sea-ice-sensitivity (CanESM5 `siconc`)

**CMORCheckError on vertical coordinate metadata (1 diagnostic):**
`alevel` coordinate's `standard_name` check fails for the `cli` variable.
- cloud-scatterplots-cli-ta (CESM2 `cli`)

### Disabled (1)

- climate-at-global-warming-levels — needs scenario data, not yet available in CMIP7

packages/climate-ref-esmvaltool/tests/unit/diagnostics/recipes/recipe_climate_at_global_warming_levels_cmip6.yml
```diff
@@ -19,6 +19,7 @@ preprocessors:
       operator: mean
     annual_statistics:
       operator: mean
+      keep_group_coordinates: true
```
lewisjared: @bouweandela Did the default for this change recently?

lewisjared:
I was trying to avoid the dependency on esmvalcore unless truely needed. The conversions don't just happen in a single standalone script.

PR #539, opened 2026-02-16
Replace out_name by variable_id in CMIP7 paths

bouweandela:
## Description

Replace "out_name" by "variable_id" in paths, as described in https://zenodo.org/records/17250297. Quoting the description of variable_id from the document:

> The variable_id in CMIP7 is also referred to as
"root name" and "out name". Usually all
variables associated with a particular
"standard name", as defined by the CF
conventions, are all assigned a single
variable_id, but sometimes the
branding_suffix cannot distinguish between
variables sharing the same standard_name, so
the variable_id may differ.

If I've understood it correctly, `out_name` is the name of the primary variable inside the file, while `variable_id` is the variable name used in the filename.

## Checklist

- [x] Tests added

### DIFF
--- a/packages/climate-ref-core/src/climate_ref_core/cmip6_to_cmip7.py
+++ b/packages/climate-ref-core/src/climate_ref_core/cmip6_to_cmip7.py
@@ -182,7 +182,7 @@ def get_cmip7_compound_name(table_id: str, variable_id: str) -> str:
     Get the full CMIP7 compound name for a CMIP6 variable.
 
     The CMIP7 compound name has the format:
-    ``<realm>.<out_name>.<branding_suffix>.<frequency>.<region>``
+    ``<realm>.<variable_id>.<branding_suffix>.<frequency>.<region>``
 
     Parameters
     ----------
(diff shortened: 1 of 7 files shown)


packages/climate-ref-core/src/climate_ref_core/cmip6_to_cmip7.py
```diff
@@ -483,14 +481,19 @@ def convert_cmip6_dataset(
...
     if variable_id is None and data_vars:
         variable_id = data_vars[0]
     if variable_id is None:  # pragma: no cover
         raise ValueError("Cannot determine variable_id for branding.")
 
     table_id = ds.attrs["table_id"]
+
+    # Use the CMIP7 variable name
+    dreq_entry = get_dreq_entry(table_id, variable_id)
+    ds = ds.rename({data_vars[0]: dreq_entry.out_name})
```
lewisjared: In a followup PR I'll replace out_name with the `physical_parameter_name`

lewisjared:
I've got a followup underway removing all use of out_name

PR #547, opened 2026-02-18
Remove the use of out_name 

lewisjared:
## Description

- Simplified the extraction of variable mappings from the CMIP7 Data Request
- Removed out_name usages
- Don't fallback to defaults when defining filenames and drs

## Checklist

- [x] Tests added

### DIFF
--- a/packages/climate-ref-core/src/climate_ref_core/cmip6_to_cmip7.py
+++ b/packages/climate-ref-core/src/climate_ref_core/cmip6_to_cmip7.py
@@ -526,7 +526,7 @@ def convert_cmip6_dataset(
 
     # Use the CMIP7 variable name
     dreq_entry = get_dreq_entry(table_id, variable_id)
-    ds = ds.rename({data_vars[0]: dreq_entry.variable_id})
+    ds = ds.rename({variable_id: dreq_entry.variable_id})
 
     # Convert global attributes
     ds.attrs = convert_cmip6_to_cmip7_attrs(ds.attrs)


packages/climate-ref-core/src/climate_ref_core/cmip6_to_cmip7.py
```diff
@@ -491,11 +454,10 @@ def convert_cmip6_dataset(
 
     # Use the CMIP7 variable name
     dreq_entry = get_dreq_entry(table_id, variable_id)
-    ds = ds.rename({data_vars[0]: dreq_entry.out_name})
+    ds = ds.rename({data_vars[0]: dreq_entry.physical_parameter_name})
```
bouweandela: I CMORized CMIP6 `tasmax` as CMIP7 `tas_tmaxavg-h2m-hxy-u` with CMOR as described in https://github.com/WCRP-CMIP/cmip7-cmor-tables/blob/main/Simple_recmorise_cmip6-cmip7.ipynb and ended up with a variable called `tas` inside the file (and not `tasmax`). Therefore, I still think that `out_name` is used for the variable name inside the file.

packages/climate-ref-core/src/climate_ref_core/cmip6_to_cmip7.py
```diff
@@ -367,62 +342,48 @@ def convert_cmip6_to_cmip7_attrs(
...
-    if "initialization_index" in attrs:
-        attrs["initialization_index"] = convert_variant_index(attrs["initialization_index"], "i")
-    if "physics_index" in attrs:
-        attrs["physics_index"] = convert_variant_index(attrs["physics_index"], "p")
-    if "forcing_index" in attrs:
-        attrs["forcing_index"] = convert_variant_index(attrs["forcing_index"], "f")
+    attrs["realization_index"] = convert_variant_index(attrs.get("realization_index", 1), "r")
+    attrs["initialization_index"] = convert_variant_index(attrs.get("initialization_index", 1), "i")
+    attrs["physics_index"] = convert_variant_index(attrs.get("physics_index", 1), "p")
+    attrs["forcing_index"] = convert_variant_index(attrs.get("forcing_index", 1), "f")
```
bouweandela: Somehow my data always seems to end up with `f1`, even when the input data has `f2`

lewisjared: Fall back to using value parsed from variant label, but it could also because `convert_variant_index` explicitly checks for `int` so numpy ints fail. Either way its more defensive now

PR #548, opened 2026-02-19
ilamb3 updates

nocollier:
## Description

- bumps the version of `ilamb3` to `v2026.2.19`.
- removes the ohc-noaa comparison as per the MBTT via Forrest
- adds a `evspsbl-pr` metric for land as per the MBTT, will need to put this [file](https://www.ilamb.org/ILAMB-Data/DATA/evspsbl/GLEAMv3.3a/et.nc) on the S3 bucket `ilamb/evspsbl/GLEAMv3.3a/et.nc sha1:5aaf73949af8c6b509ef16f684aa8efeccd983e2`
- uses obs4REF data when possible. Cannot use WOA data as its date range starts after CMIP6 historical ends. Cannot use RAPID as someone removed dimensional information (depth_bonds, latitude) that I need.
- fixes model line plots being indistinguishable from the reference
- all ilamb/iomb are running for me locally
 
## Checklist

- [ ] Tests added

### DIFF
--- a/packages/climate-ref-ilamb/src/climate_ref_ilamb/configure/ilamb.yaml
+++ b/packages/climate-ref-ilamb/src/climate_ref_ilamb/configure/ilamb.yaml
@@ -16,6 +16,7 @@ gpp-FLUXNET2015:
 mrro-LORA:
   sources:
     mrro:
+      obs_source: obs4ref
       source_id: LORA-1-1
   variable_cmap: Blues
 
@@ -38,6 +39,7 @@ cSoil-HWSD2:
 lai-NOAA-NCEI-LAI-AVHRR-5-0:
   sources:
     lai:
+      obs_source: obs4mips
       source_id: NOAA-NCEI-LAI-AVHRR-5-0
   variable_cmap: Greens
 
@@ -46,6 +48,7 @@ nbp-Hoffman:
   - nbp
   sources:
     nbp:
+      obs_source: obs4ref
       variable_id: nbp
       source_id: Hoffman
   transforms:
--- a/packages/climate-ref-ilamb/src/climate_ref_ilamb/standard.py
+++ b/packages/climate-ref-ilamb/src/climate_ref_ilamb/standard.py
@@ -1,4 +1,4 @@
-from collections.abc import Sequence
+from collections.abc import Mapping, Sequence
 from pathlib import Path
 from typing import Any
 
@@ -21,7 +21,8 @@ from climate_ref_core.diagnostics import (
     ExecutionDefinition,
     ExecutionResult,
 )
-from climate_ref_core.esgf import CMIP6Request, CMIP7Request
+from climate_ref_core.esgf import CMIP6Request, CMIP7Request, RegistryRequest
+from climate_ref_core.esgf.obs4mips import Obs4MIPsRequest
 from climate_ref_core.metric_values.typing import SeriesMetricValue
 from climate_ref_core.pycmec.metric import CMECMetric
 from climate_ref_core.pycmec.output import CMECOutput, OutputCV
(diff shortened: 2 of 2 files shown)


packages/climate-ref-ilamb/src/climate_ref_ilamb/configure/ilamb.yaml
```diff
@@ -2,10 +2,9 @@ registry: ilamb
 
 gpp-WECANN:
   sources:
-    # TODO: Update to use the obs4REF equiv
-    gpp: ilamb/gpp/WECANN/gpp.nc
+    gpp: obs4REF/ColumbiaU/WECANN-1-0/mon/gpp/gn/20250516/gpp_mon_WECANN-1-0_REF_gn_200701-201512.nc
   relationships:
-    pr: ilamb/pr/GPCPv2.3/pr.nc
+    pr: obs4REF/NOAA-NCEI/GPCP-2-3/mon/pr/gn/v20210727/pr_mon_GPCP-2-3_PCMDI_gn_197901-201907.nc
```
lewisjared: Can you move to using a data requirement for these if they are obs4REF data instead of hardcoding the paths?

https://github.com/Climate-REF/climate-ref/blob/853097eb2e5b47d752735903285d64793ac28ce6/packages/climate-ref-pmp/src/climate_ref_pmp/diagnostics/enso.py#L211

packages/climate-ref-ilamb/src/climate_ref_ilamb/configure/iomb.yaml
```diff
@@ -30,20 +30,5 @@ amoc-RAPID:
   transforms:
   - msftmz_to_rapid
   sources:
-    # TODO: Update to use the obs4REF equiv
+    # obs4REF equiv does not work, changes `depth` to `olevel` and removed the depth bounds
```
lewisjared: @Darwyn72

packages/climate-ref-ilamb/src/climate_ref_ilamb/configure/ilamb.yaml
lewisjared: Ahh I see your comment below. 

Can we add the data requirements anyway and ignore them because I'd like to use the information in the database/docs

nocollier: Yeah, I feel you there. In my mind I was going to try that in a followup PR but I can do it here. This is what I had in my mind to do:

1. Look at the ilamb configure block's sources and if they contain obs4MIPs or obs4REF, parse out some search criteria from the key to create a DataRequirement. 
2. If successful, then we insert this into the requirements
3. If not successful, then we fall back to what I am already doing. I just have to figure out how not to use registries if I have a obs4whatever datarequirement.

I haven't tried it yet, but do you think this approach will work?

packages/climate-ref-ilamb/src/climate_ref_ilamb/configure/iomb.yaml
```diff
@@ -2,7 +2,7 @@ registry: iomb
 
 thetao-WOA2023-surface:
   sources:
-    # TODO: Update to use the obs4REF equiv
+    # NOTE: obs4REF equiv starts beyond historical, cannot use for CMIP6
```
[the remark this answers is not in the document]
Darwyn72: @lewisjared This is the [latest list](https://www.ilamb.org/ilamb3-data/) of ILAMB prepared datasets. Morgan Steckler asks that any errors found by the REF with the data are logged [here](https://github.com/rubisco-sfa/ilamb3-data) I can't forward the data to you by email as the files are too large. For WOA2023 the [two files](https://www.ilamb.org/ilamb3-data/WOA-23/) for thetao and so are dated 2025-10-24. Thanks, Paul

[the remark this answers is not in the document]
lewisjared: Will this be the official location of these data? Will these be moved to CEDA?

I'm a bit wary as I'm pretty sure the REF isn't using all of the latests datasets you have curated. Is there a complete list of download URLs for all datasets that have been vetted?

nocollier:
Thanks for the help @lewisjared. I have pushed another change which makes use of obs4MIPs Data Requirements if a dictionary of keywords is given in the configure. In the end I spent a while trying to figure out how to make `mypy` not complain about a few lines and just pushed them here to see if you have a suggestion.

I was asked to use the lai dataset that has been ingested into obs4MIPs already and I did change the configure to use it. However, they have ingested a 5km version and it won't run on my machine without running out of memory. It should work--the others do--but I didn't get to look at the output.

lewisjared:
@nocollier I merged main back into this branch and fixed the mypy errors. The new dataset is also uploaded.

This will require adding extra datasets to the data catalogs we use to test if the constraints are working. Currently we can only generate these from scratch. 

I'll have a look tomorrow to see how we can unblock that.

Issue #545, opened 2026-02-18
Select only one version of each dataset


bouweandela:
## The problem

There appear to be two versions of the same dataset selected for the climate drivers for fire diagnostic:
https://github.com/Climate-REF/climate-ref/blob/5c0817f6b31eedea9ed6109e99cd65f9fb6cba5a/packages/climate-ref-esmvaltool/tests/unit/test_solve_regression/test_solve_regression_climate_drivers_for_fire_.yml#L96-L97

## Definition of "done"

Only one version of each dataset should be included in the result of a solve. Maybe only the latest version, so `CMIP6.CMIP.EC-Earth-Consortium.EC-Earth3-Veg.historical.r1i1p1f1.Amon.tas.gr.v20211207` and not `CMIP6.CMIP.EC-Earth-Consortium.EC-Earth3-Veg.historical.r1i1p1f1.Amon.tas.gr.v20200225`

Issue #543, opened 2026-02-18
Additional experiments appearing in the global warming level diagnostic


lewisjared:
## The problem

https://github.com/Climate-REF/climate-ref/blob/main/packages/climate-ref-esmvaltool/tests/unit/test_solve_regression/test_solve_regression_climate_at_global_warming_levels_.yml

## Definition of "done"

solve regression test only has historical and target experiment

PR #554, opened 2026-02-20
fix: apply version filtering and fix supplementary dataset selection in regression analysis

lewisjared:
## Description

Fixes two issues in the dataset selection logic that caused spurious entries in regression analysis results:

1. **Version filtering in dataset queries** (`datasets/base.py`): Added proper `version` filtering to `query_datasets` and `query_facets` methods. Previously, version constraints were silently ignored, causing older dataset versions to leak into results. The version filter supports both exact matching and ordering comparisons (e.g. `v20190731` vs `v20200101`).

2. **Supplementary dataset selection** (`constraints.py`): Fixed `AddSupplementaryDataset` to pick a single best-matching dataset per group instead of accumulating multiple candidates across score ties. Previously, when multiple supplementary datasets tied on matching score, all were selected, leading to spurious experiment entries (e.g. `1pctCO2`, `esm-1pct-brch-1000PgC`) appearing alongside expected `historical` and SSP data.

These fixes together eliminate unexpected experiment/version combinations from regression outputs.

## Checklist

- [x] Tests added

### DIFF
--- a/changelog/553.fix.md
+++ b/changelog/553.fix.md
@@ -1,1 +1,0 @@
-Fixed supplementary dataset selection to pick a single dataset per matching group instead of accumulating multiple candidates across score ties. This eliminates spurious experiment entries (e.g. `1pctCO2`, `esm-1pct-brch-1000PgC`) from appearing alongside expected `historical` and SSP data.
--- a/changelog/554.fix.md
+++ b/changelog/554.fix.md
@@ -1,0 +1,1 @@
+Applied version filtering to dataset queries and fixed supplementary dataset selection to pick a single dataset per matching group. This eliminates spurious experiment and version entries from regression analysis results.


lewisjared:
@bouweandela what do you think about https://github.com/Climate-REF/climate-ref/pull/554/changes/1cb40c028f0075d91197d226617cb90c23e54ef4 ?

lewisjared:
Closes #545 and #543

bouweandela:
For reference, here is an overview of the supplementary variables used by ESMValCore preprocessor functions: https://docs.esmvaltool.org/projects/ESMValCore/en/latest/recipe/preprocessor.html#supplementary-variables-ancillary-variables-and-cell-measures. Not all preprocessor functions are used by the REF, so not all of them may be needed.

Note that I have also used the `AddSupplementary` constraint to add the `historical` experiment to the scenarios here: https://github.com/Climate-REF/climate-ref/blob/cf05fe96d8d3d811bfc011c63ae5abdf7bcf0a6a/packages/climate-ref-esmvaltool/src/climate_ref_esmvaltool/diagnostics/climate_at_global_warming_levels.py#L72-L76. Those variables will certainly be time-dependent.

packages/climate-ref-core/src/climate_ref_core/constraints.py
```diff
@@ -248,16 +248,19 @@ def apply(
             facets = matching_facets + list(self.optional_matching_facets)
             datasets = group[facets].drop_duplicates()
             select = pd.Series(False, index=supplementary_group.index)
-            for i in range(len(datasets)):
-                dataset = datasets.iloc[i]
-                # Restrict the supplementary datasets to those that match the main dataset.
+            for _, main_group_df in datasets.groupby(matching_facets):
```
bouweandela: I tried to understand what is happening here, but this is getting rather complicated. Will have another look tomorrow.

PR #555, opened 2026-02-20
ci: use ARC runners for integration tests

lewisjared:
## Description

Migrate the integration test workflow from generic `self-hosted` runners to the `arc-climate-ref` ARC runner scale set. This involves three changes:

- **Use `arc-climate-ref` runner label** for all integration test jobs, matching the ARC runner scale set registered at the org level
- **Replace `make` calls with direct `uv` commands** since the ARC runner image doesn't have `make` installed (`make virtual-environment` -> `uv sync`, `make fetch-test-data` -> direct `uv run ref datasets` commands)
- **Add a `populate-cache` job** that runs before the test jobs to download datasets, create conda environments, and fetch test cases on the shared NFS volume, avoiding race conditions from concurrent writes

## Checklist

- [ ] Tests added


lewisjared:
I think the integration tests are still failing, but that is a separate issue

PR #565, opened 2026-02-23
fix: pandas 3 compatibility

lewisjared:
## Description

Resolve pandas FutureWarnings and compatibility issues to support both pandas 2 and 3:

- **`constraints.py`**: Skip `pd.concat` when `supplementary_group` is empty; drop all-NA columns before concat to avoid dtype determination changes in pandas 3
- **`example.py`**: Drop `time_bnds` before weighted mean (not after) to prevent `dtype=object` division error from `resample+mean` on datetime bounds; set explicit `compat="override"` on `open_mfdataset`
- **`test_solve_helpers.py`**: Normalize `None`/`NaT` nulls in parquet round-trip assertions to handle stricter equality checks in pandas 3
- Upgraded lockfile to the latest dependencies

## Checklist

- [x] Tests added

### DIFF
--- a/packages/climate-ref-ilamb/pyproject.toml
+++ b/packages/climate-ref-ilamb/pyproject.toml
@@ -24,10 +24,9 @@ classifiers = [
 ]
 dependencies = [
   "climate-ref-core",
-  "ilamb3>=2025.9.9,<2025.12",  # >=2025.12 removed integrate_space from nbp_analysis
+  "ilamb3>=2025.9.9,<2025.12",  # to be relaxed in #548
   "scipy<1.16",        # https://github.com/statsmodels/statsmodels/issues/9584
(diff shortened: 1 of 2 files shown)


packages/climate-ref-ilamb/pyproject.toml
```diff
@@ -24,7 +24,7 @@ classifiers = [
 ]
 dependencies = [
   "climate-ref-core",
-  "ilamb3>=2025.9.9",
+  "ilamb3>=2025.9.9,<2025.12",  # to be relaxed in #548
```
lewisjared: @nocollier This will cause a merge conflict with #548, but we can resolve when merging

lewisjared:
Pre-req for #506

PR #582, opened 2026-03-06
Fetch ESMValTool recipes in advance

### LINKED ISSUE #468: ESMValTool requires access to the web to retrieve the recipes
https://github.com/Climate-REF/climate-ref/blob/01eabb53ecf47caf5d42f69e3e3601ac002c33ff/packages/climate-ref-esmvaltool/src/climate_ref_esmvaltool/recipe.py#L150C12-L150C17

Hi,

I have been running some tests on the HPC and ran into this issue today. It is a bit of a problem because, on many HPC systems, internet access is restricted to specific queues with very limited resources. It also feels sub-optimal in general. The recipes are already shipped with the ESMValTool package, so I would expect them to be retrieved from the installed package rather than fetched online.

An alternative could be to support an environment variable that points to one or more local paths to search for recipes.

@bouweandela, keen to hear your thoughts. At the moment this is quite limiting for us, and it means we cannot use REF on Gadi.

bouweandela:
## Description

Closes #468

## Checklist

- [x] Tests added


lewisjared (approved):

lewisjared (commented):
The change in directory will cause users to need to redownload the data again. Informing users that a refetch (`ref providers setup --provider esmvaltool`) might be easier than trying to migrate the old cache. What do you think?

lewisjared:
It wasn't too bad to add a migration. This was required for the ilamb/pmp/sample-data registries.

The added benefit is that only the actively used files were moved to the right directory making it easy to clean up dead files

PR #636, opened 2026-04-27
Update CITATION.cff with extra authors and changes to spelling

Darwyn72:
Updated citation file to version 1.2.1 and added new authors with ORCID links. Updated accent on Dora Hegedūs name.

## Checklist

- [ ] Tests added

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,4 +1,4 @@
-cff-version: 1.2.0
+cff-version: 1.2.1
 message: "If you use this software, please cite it as below."
 title: "Rapid Evaluation Framework"
 abstract: "An extensible framework for the evaluation of Earth System Models."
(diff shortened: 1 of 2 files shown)


CITATION.cff
```diff
@@ -1,4 +1,4 @@
-cff-version: 1.2.0
+cff-version: 1.2.1
```
lewisjared:
```suggestion
cff-version: 1.2.0
```
This refers to the version of the [Citation File Format](https://citation-file-format.github.io/). There is no such thing as v1.2.1.

lewisjared (changes requested):
Thanks.

Can you also add a changelog entry. See: https://github.com/Climate-REF/climate-ref/tree/main/changelog

The target of this PR should be the `main` branch (I updated it)

CITATION.cff
Darwyn72: @lewisjared I think I've added the changelog correctly it's 636.dos.md - can you confirm?

Darwyn72:
Added a changelog - 636.docs.md to main

lewisjared:
Ahh I meant that the target for this branch should be main, not to commit the changelog to main. I'll revert and add back to this branch

image

lewisjared (approved):
I undid the commit on main and added the changelog on this branch

Issue #572, opened 2026-02-27
HPCExecutor configurations


zmaalick:
Hi, 

We are trying to run HPCExecutor.

Here are our configurations:

```
[executor]
executor = "climate_ref.executor.HPCExecutor"

[executor.config]
address = "eth0"
address_probe_timeout = 30
scheduler = "slurm"
account = "normal" # our account name
username = "my_username"
partition = "cpu"           # or "debug" for testing
#qos = "debug"                   # quality of service
req_nodes = 1                   # number of nodes
walltime = "00:30:00"           # time limit per job
cores_per_worker = 1            # cores per worker
max_workers_per_node = 4       # workers per node
scheduler_options = ""
worker_init = "export PARSL_HTEX_ADVERTISED_ADDRESS=10.138.80.53; export PARSL_HTEX_ADDRESS=10.138.80.53;"
```

but we are getting connection error:

```
0:     raise ConnectionError(f"No viable ZMQ url from: {addys}")
0: ConnectionError: No viable ZMQ url from: tcp://127.0.0.1:54201, tcp://10.138.80.53:54201, tcp://151.170.240.197:54201
srun: error: sla-cpu-r-03: task 0: Exited with exit code 1
```

I above settings we did a bit of work around by adding scheduler_options but didn't work. Anything else we need in configurations?

minxu74:
@zmaalick Sorry for the long to fix your issue. If you have time, could you help test my branch in the PR #651? You could add the additional options of parsl executor or provider in the `ref.toml` using:

```
[executor.config.extra_slurm_provider]
option1 = vaule1

[executor.config.extra_executor]
option1 = vaule1
```

minxu74:
Closed as completed

PR #651, opened 2026-05-02
add executor.config.extra_* so options other than base can be injected to parsl functions

minxu74:
## Description
Currently, `HPCExecutor` only provided limited parsl options (base options) to users, in some HPCs, full sets of options of parsl functions (`SlurmProvider` and `HighThroughputExecutor`) are needed. This PR provides two dictionaries `extra_slurm_provider` and `extra_executor` to let users specify the parsl options. These two dicts are set in `ref/toml` under [executor.config.extra_slum_provider] and [executor.config.extra_executor]. The extra options won't override the base options including their defaults. To change the base options, users have to set them explicitly under [executor.config].
## Checklist

- [x] Tests added

Fixes: issue #572

### DIFF
--- a/changelog/651.improvement.md
+++ b/changelog/651.improvement.md
@@ -1,0 +1,1 @@
+Add extra options for users to ingest options other than base to parsl functions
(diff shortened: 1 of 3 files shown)


minxu74:
@lewisjared the pr is ready for your review. I need to add a how-to to the doc.

lewisjared (approved):

Issue #647, opened 2026-04-30
Improve CMIP7 support for diagnostics that require a parent dataset


bouweandela:
There are several issues with the current CMIP7 support of diagnostics that require a parent dataset:

## ECS

The `update_recipe` method is not using the function `get_child_and_parent_dataset` for CMIP7 data, likely resulting in a wrong recipe:
https://github.com/Climate-REF/climate-ref/blob/9f5541a11999269946dc1d7b9610025b79a6efbc/packages/climate-ref-esmvaltool/src/climate_ref_esmvaltool/diagnostics/ecs.py#L213-L236

## TCR and TCRE

The TCR and TCRE data requirements do not use the constraint `AddParentDataset`, likely resulting in the wrong input data being selected:
https://github.com/Climate-REF/climate-ref/blob/9f5541a11999269946dc1d7b9610025b79a6efbc/packages/climate-ref-esmvaltool/src/climate_ref_esmvaltool/diagnostics/tcr.py#L42-L84
https://github.com/Climate-REF/climate-ref/blob/9f5541a11999269946dc1d7b9610025b79a6efbc/packages/climate-ref-esmvaltool/src/climate_ref_esmvaltool/diagnostics/tcre.py#L77-L109

schlunma:
I am struggling to set this up properly for TCRE. For this, we need 3 data sets per model run:

- `tas` from the `esm-flat10` experiment (for CMIP6: `esm-1pctCO2`)
- `tas` from the parent of the `esm-flat10` experiment (i.e, `esm-piControl`)
- `fco2antt` from the `esm-flat10` experiment

Currently, we have the following for CMIP6

```py
DataRequirement(
                source_type=SourceDatasetType.CMIP6,
                filters=(
                    FacetFilter(
                        facets={
                            "variable_id": "tas",
                            "experiment_id": "esm-1pctCO2",
                            "table_id": "Amon",
                        },
                    ),
                ),
                group_by=("source_id", "member_id", "grid_label"),
                constraints=(
                    AddParentDataset.from_defaults(SourceDatasetType.CMIP6),
                    AddSupplementaryDataset(
                        supplementary_facets={
                            "variable_id": "fco2antt",
                            "experiment_id": "esm-1pctCO2",
                        },
                        matching_facets=(
                            "source_id",
                            "member_id",
                            "table_id",
                            "grid_label",
                        ),
                        optional_matching_facets=("version",),
                    ),
[...]

PR #686, opened 2026-05-22
Make TCR diagnostic compatible with CMIP7 data

schlunma:
## Description

Addresses https://github.com/Climate-REF/climate-ref/issues/647.

## Checklist

- [x] Tests added

## Summary by CodeRabbit

* **New Features**
  * The TCR diagnostic now supports CMIP7 data.
  * Updated the diagnostic workflow to handle the revised CMIP7 dataset setup.

* **Bug Fixes**
  * Improved handling of dataset branding details so results generate correctly for both CMIP6 and CMIP7.
  * Expanded test coverage and sample data to reflect additional supported model variants.


packages/climate-ref-esmvaltool/src/climate_ref_esmvaltool/diagnostics/tcr.py
```diff
@@ -67,16 +66,16 @@ class TransientClimateResponse(ESMValToolDiagnostic):
...
+                            "experiment_id": "1pctCO2",
                             "frequency": "mon",
                             "region": "glb",
                         },
                     ),
                 ),
                 group_by=("source_id", "variant_label", "grid_label"),
                 constraints=(
+                    AddParentDataset.from_defaults(SourceDatasetType.CMIP7),
                     RequireContiguousTimerange(group_by=("instance_id",)),
```
schlunma, line 74: Question on this: What does `RequireContiguousTimerange` exactly do here? Does it ensure that time ranges are contiguous for all elements within a group? Here, since groups are formed with the `instance_id`, the groups should only contain a single element each, which would make this line redundant.

schlunma:
Will mark this as "ready for review" once corresponding ESMValTool PR has been merged.

packages/climate-ref-esmvaltool/src/climate_ref_esmvaltool/diagnostics/tcr.py
bouweandela: > Does it ensure that time ranges are contiguous for all elements within a group?

Yes. A single `instance_id` may consist of multiple files, e.g. one file per year or one file per decade, so this will only trigger an execution when all files covering the required timerange have been published.

schlunma: I see, makes sense. What about `RequireOverlappingTimerange(group_by=("instance_id",))`? This is used for the ECS diagnostics, but not here. Would it make sense to add it here or remove it for the ECS diagnostic?

bouweandela: The ECS diagnostic uses 4 input variables, so for ECS that constraint makes sure that there is at least some time period for which all 4 variables are available. This diagnostic only uses 1 variable, so adding that constraint here wouldn't add much because 1 variable always overlaps with itself.

bouweandela: Data selection would be simpler here (and for ECS) if we knew the timerange in advance, but as far as I understand we don't, so these checks are just there to remove the most obvious cases in which the recipe will fail to run.

schlunma: > The ECS diagnostic uses 4 input variables, so for ECS that constraint makes sure that there is at least some time period for which all 4 variables are available. This diagnostic only uses 1 variable, so adding that constraint here wouldn't add much because 1 variable always overlaps with itself.

That makes sense. The thing that's puzzling me is that it uses `group_by=("instance_id",)`. As far as I know, the variable name is included in the instance ID, so you should end up with one data set per group (8 groups in total = 4 variables x 2 experiments). Does the check for overlapping time ranges work across groups or within groups? I would think within groups, but in this case it should not filter out anything.

> Data selection would be simpler here (and for ECS) if we knew the timerange in advance, but as far as I understand we don't, so these checks are just there to remove the most obvious cases in which the recipe will fail to run.

Yes, we do not know the time ranges in advance.

bouweandela: The constraints are applied sequentially, so in the ECS case:
https://github.com/Climate-REF/climate-ref/blob/9e7c5c056e165aded8105731a9658cec7a96e2dc/packages/climate-ref-esmvaltool/src/climate_ref_esmvaltool/diagnostics/ecs.py#L96-L104
the constraint `RequireOverlappingTimerange(group_by=("instance_id",))` results in at most 4 groups with the child experiment only, depending on data availability. Only if these groups have overlapping timerange, the next step `AddParentDataset` is applied.

schlunma: I see, makes sense. Thanks!

packages/climate-ref-esmvaltool/tests/unit/test_solve_regression/test_solve_regression_transient_climate_response_.yml
```diff
@@ -449,6 +449,10 @@ cmip7_gn_CAMS-CSM1-0_r1i1p1f1:
   cmip7:
   - CMIP7.CMIP.CAMS.CAMS-CSM1-0.1pctCO2.r1i1p1f1.glb.mon.tas.tavg-h2m-hxy-u.gn.v20190708
   - CMIP7.CMIP.CAMS.CAMS-CSM1-0.piControl.r1i1p1f1.glb.mon.tas.tavg-h2m-hxy-u.gn.v20190729
+cmip7_gn_CAMS-CSM1-0_r2i1p1f1:
+  cmip7:
+  - CMIP7.CMIP.CAMS.CAMS-CSM1-0.1pctCO2.r2i1p1f1.glb.mon.tas.tavg-h2m-hxy-u.gn.v20190726
+  - CMIP7.CMIP.CAMS.CAMS-CSM1-0.piControl.r1i1p1f1.glb.mon.tas.tavg-h2m-hxy-u.gn.v20190729
```
schlunma, line 455: From the code, it looks like `AddParentDataset` considers the attribute `parent_variant_label`. If yes, this should be ok. Can you confirm?

packages/climate-ref-esmvaltool/tests/unit/test_solve_regression/test_solve_regression_transient_climate_response_.yml
```diff
@@ -516,15 +520,55 @@ cmip7_gn_CanESM5_r1i1p2f1:
...
+  cmip7:
+  - CMIP7.CMIP.CCCma.CanESM5.1pctCO2.r1i1p1f1.glb.fx.areacella.ti-u-hxy-u.gn.v20190429
+  - CMIP7.CMIP.CCCma.CanESM5.1pctCO2.r2i1p1f1.glb.mon.tas.tavg-h2m-hxy-u.gn.v20190429
+  - CMIP7.CMIP.CCCma.CanESM5.piControl.r1i1p1f1.glb.mon.tas.tavg-h2m-hxy-u.gn.v20190429
+cmip7_gn_CanESM5_r2i1p2f1:
+  cmip7:
+  - CMIP7.CMIP.CCCma.CanESM5.1pctCO2.r1i1p1f1.glb.fx.areacella.ti-u-hxy-u.gn.v20190429
+  - CMIP7.CMIP.CCCma.CanESM5.1pctCO2.r2i1p2f1.glb.mon.tas.tavg-h2m-hxy-u.gn.v20190429
+  - CMIP7.CMIP.CCCma.CanESM5.historical.r1i1p2f1.glb.fx.areacella.ti-u-hxy-u.gn.v20190429
+  - CMIP7.CMIP.CCCma.CanESM5.piControl.r1i1p2f1.glb.mon.tas.tavg-h2m-hxy-u.gn.v20190429
```
schlunma, line 533: I am not quite sure why two different `areacella` files are found

packages/climate-ref-esmvaltool/tests/unit/test_solve_regression/test_solve_regression_transient_climate_response_.yml
```diff
@@ -612,6 +676,9 @@ cmip7_gn_MRI-ESM2-0_r1i2p1f1:
   - CMIP7.CMIP.MRI.MRI-ESM2-0.1pctCO2.r1i1p1f1.glb.fx.areacella.ti-u-hxy-u.gn.v20190603
   - CMIP7.CMIP.MRI.MRI-ESM2-0.1pctCO2.r1i2p1f1.glb.mon.tas.tavg-h2m-hxy-u.gn.v20191205
   - CMIP7.CMIP.MRI.MRI-ESM2-0.piControl.r1i2p1f1.glb.mon.tas.tavg-h2m-hxy-u.gn.v20191205
+cmip7_gn_NESM3_r1i1p1f1:
+  cmip7:
+  - CMIP7.CMIP.NUIST.NESM3.1pctCO2.r1i1p1f1.glb.mon.tas.tavg-h2m-hxy-u.gn.v20190707
```
schlunma, line 681: This looks wrong, there's no parent experiment.

schlunma:
I copied the obtained file from the failing regression test to make it pass here. However, I think there are some problems with this (see my comments).

Otherwise this is ready for review now.

packages/climate-ref-esmvaltool/tests/unit/test_solve_regression/test_solve_regression_transient_climate_response_.yml
bouweandela: Yes, but I noticed some other facets are missing. Opened #701 to report it.

packages/climate-ref-esmvaltool/tests/unit/test_solve_regression/test_solve_regression_transient_climate_response_.yml
bouweandela: Because we use the closest match, which may be different for different experiments. It shouldn't be a problem because they are all the same anyway.

schlunma: Got it! As long as this doesn't lead to concatenation errors, this should be fine.

packages/climate-ref-esmvaltool/tests/unit/test_solve_regression/test_solve_regression_transient_climate_response_.yml
lewisjared: I'll fix in a follow up. Its due to RequireContiguousTimerange being too sensitive.

lewisjared (approved):

PR #744, opened 2026-06-19
feat(regression): stabilise committed-bundle bytes, fix docs & lint coverage

lewisjared:
## Description

Three orthogonal core/tooling improvements surfaced while doing the ILAMB cmip6 baseline work (handoff from #743). Each is cross-cutting — it affects every provider's regression bundle — so they are grouped here as one small PR with separate commits. None re-runs or re-mints any baselines.

**1. Committed-bundle float quantisation (`feat`).** Committed `series.json` / `diagnostic.json` stored full-precision `float64` whose least-significant digits are platform-dependent (CPU, BLAS, library versions), so they churned byte-for-byte between a local `run` and a CI `mint` even when numerically identical. Floats are now rounded to **7 significant figures** at write time, in core, so it applies to all providers and to both `run` and `mint`. Seven figures is deliberately one digit finer than the `rtol=1e-6` regression compare, so the rounding error stays an order of magnitude under tolerance and can never flip a gate verdict. `output.json` is float-free and left untouched; native blobs and their content-addressed digests are never touched. This is **non-breaking without a re-mint**: the gate compares values at `rtol=1e-6`, so 7-s.f. output still matches the existing full-precision baselines — a one-time re-mint to refresh the stored bytes can follow separately whenever convenient.

**2. Documentation fixes (`docs`).** Corrected inaccuracies in `testing-diagnostics.md` (a non-existent `TestCase.datasets` attribute and three-level resolution priority; the size-flagging attribution, which is `ref test-cases run --size-threshold`, not a pre-commit hook; and the `execution_regression` factory + `check()` example), and filled gaps in `regression-baselines.md` (the manifest `schema` field, the `check-store` verb, `CMIP7Request` and the `cmip7-converted` cache, and a note on the new committed-bundle float precision).

**3. Pre-commit exclude fix (`ci`).** The `.*/regression/.*` exclude — intended to skip the committed baseline *data* — also matched the regression *source* package (`climate_ref_core/regression/`) and its unit tests, so ruff never linted them (CI runs only `make pre-commit`). The pattern is narrowed to data under `test-data/` locations so the source is linted like any other code while the baseline bundles stay excluded; zero pre-existing errors surfaced.

### Verification

38 new + 191 existing regression unit tests pass; `ruff`, `ruff format` and `mypy` are clean; the full `make pre-commit` run is green and modifies no files.

## Checklist

- [x] Tests added

## Summary by CodeRabbit

* **New Features**
  * Regression baseline outputs now deterministically round floating-point values to **7 significant figures**, producing stable committed bundle bytes and matching digests across environments.

* **Bug Fixes**
  * Committed regression digests now correspond to the rounded JSON bytes to prevent digest mismatches.

* **Documentation**
  * Updated regression baseline and diagnostic testing guides with clarified schemas, new `check-store` lifecycle guidance, dataset resolution workflow changes, and float precision details.

* **Chores**
  * Refined pre-commit exclusions to skip only committed regression baseline data under `test-data/`, and enabled linting for the regression source and tests.

* **Tests**
  * Added unit tests covering float quantisation and rounded committed bundle output.


lewisjared:
@coderabbitai

PR #791, opened 2026-07-06
fix: let read-only consumers build the provider registry without writing

lewisjared:
## Summary

`ProviderRegistry.build_from_config` unconditionally upserts provider and diagnostic rows via `_register_provider`, which is the only step in that method that writes to the database. That makes it impossible for a read-only consumer — notably the `ref-app` API serving a database mounted read-only — to build the registry: startup crashes with `attempt to write a readonly database` even though the rows it needs are already registered by `ref providers setup` and the workers.

This adds a `register` argument to `build_from_config` (default `True`, so every existing caller is unchanged). Read-only consumers pass `register=False` to hydrate the registry from imported providers without touching the database.

## Changes

- `ProviderRegistry.build_from_config` gains `register: bool = True`; the `_register_provider` write loop is skipped when it is `False`.
- Unit test for the `register=False` path, plus a `TestProviderRegistryReadOnly` integration test that migrates and registers a real database, reopens it read-only, and asserts that `register=False` builds with no write while `register=True` raises `OperationalError` (the exact crash this fixes).

## Context

This is the `climate-ref` half of the fix for Climate-REF/ref-app#31; the `ref-app` API is updated separately to pass `register=False` on its read-only path.

## Summary by CodeRabbit

* **New Features**
  * You can now build the provider registry without writing to the database, which supports read-only environments.
  * The registry still loads provider configuration and works in memory as before.

* **Bug Fixes**
  * Improved handling for read-only databases by avoiding write transactions when registration is disabled.
  * Added coverage to confirm database writes fail as expected in read-only mode when registration is enabled.


lewisjared:
@coderabbitai review

lewisjared:
@coderabbitai review

PR #814, opened 2026-07-12
feat: ingest obs4REF and ESMValTool reference datasets

lewisjared:
## Description

Adds the adapters that make the `obs4REF` and `ESMValToolReference` types from #805 ingestible. #805 added the models and tables. This adds the ingest path.

- `Obs4REFDatasetAdapter` subclasses the obs4MIPs adapter. obs4REF shares the obs4MIPs metadata conventions, so it reuses the parser and overrides only the accepted activity ids and the `instance_id` prefix.
- `parse_obs4mips` gains an `accepted_activity_ids` hook. A file whose `activity_id` is outside the adapter's set but is still a known obs4MIPs or obs4REF id is ingested with a warning, not rejected. The hard reject for that cross-contamination case is a follow-up.
- `ESMValToolReferenceDatasetAdapter` parses ESMValTool's own layout, which is `OBS`, `OBS6`, `native6` and a non-compliant `obs4MIPs` subtree.
- It records `frequency` and no MIP table. `native6` carries a frequency directly. `OBS`/`OBS6` carry a table that `frequency_from_mip_table` reduces to one, so `Amon` becomes `mon`.

obs4REF is wired through the solver and catalog. ESMValTool reference is dispatch only for now, which matches its provenance-first use case.

One known gap. `frequency_from_mip_table` does not map `OImon`, the CMIP5-era ocean sea-ice table. Its CMIP6 successor `SImon` is mapped. An OBS6 dataset using `OImon` fails ingestion with a clear error rather than storing bad data, so this is a coverage gap, not a silent bug. Left for a real dataset to force.

Unit datasets and solver suites pass, 389 passed. ruff and mypy clean.

## Checklist

- [x] Tests added


lewisjared:
@coderabbitai review

PR #843, opened 2026-07-27
feat: fetch ESMValTool reference data for a test case

lewisjared:
Adds a `RegistryRequest` parser for the ESMValTool reference registry, so a test case can fetch the reference data an opt-in diagnostic selects. This closes the first of the two gaps #835 left behind, and is what a first adopter needs before its baseline can be minted.

- Adds an `esmvaltool-datasets` branch to `RegistryRequest._get_parser`. Without it a fetch fell through to the obs4REF parser, which rejects every key, so a test case came back with nothing to solve.
- Moves the path-to-metadata step into `climate_ref_core.esmvaltool_reference`, alongside the anchor logic that already lives there. The ingest adapter now delegates to it rather than carrying its own copy.
- Moves `frequency_from_mip_table` into the same module. Nothing outside the ESMValTool reference DRS used it.
- Normalises facet values in `_matches_facets` against a `Collection` rather than a `str`, and adds a `FacetValue` alias for the annotation. `tier` parses as an int, which the previous signature denied and the previous `tuple(value)` call would have raised a `TypeError` on.

Routing both sides through one function is the point of the refactor. A request cannot name a facet value that ingest would spell differently, so a key ending `_OImon_sic_` yields `frequency="mon"` on the fetch side and the catalog side alike. A separate regex in `registry.py` would very plausibly have yielded `OImon` and silently matched nothing.

All 165 keys in the registry parse, split 91 `OBS`, 70 `native6`, 3 `OBS6` and 1 `obs4MIPs`. No behaviour changes for the pmp-climatology or obs4ref registries.

One finding worth recording while checking the second gap #835 described. Latest-version filtering still prefers a `vN` version over one like `Ed4.2`, but no `(source_id, variable_id)` group in the registry carries more than one version, so it cannot drop anything today. That gap stays deferred.

No diagnostic declares the requirement yet, so nothing existing changes. Declaring it on `sea-ice-area-basic` is the next step.


lewisjared:
@coderabbitai review

PR #846, opened 2026-07-27
feat: ingest ESMValTool reference data at setup and declare it on sea ice

lewisjared:
ESMValTool now ingests the reference data it downloads, and `sea-ice-area-basic` is the first diagnostic to declare what it needs.

- Adds `ESMValToolProvider.ingest_data`, so `ref providers setup` catalogs the fetched registry the way the PMP provider does. Fetching alone left the solver with nothing to select from, and a user had to run `ref datasets ingest --source-type esmvaltool-reference` by hand.
- Adds an `esmvaltool-reference` data requirement to `sea-ice-area-basic`, selecting the four OSI-450 datasets the recipe compares each model against.
- Adds the matching registry request to both test cases, so a baseline can fetch the same 74 files.
- Moves the data root behind `registry_data_root`, shared by ingest and the whole-registry fallback.
- Adds the `esmvaltool_reference_catalog.parquet` solve catalog the tests need.
- Widens the registry request facets to a `Mapping`, so the solve filter and the fetch request can share one frozen dict rather than repeating it. This retires a type ignore in the ILAMB test data spec.

A diagnostic that does not declare the requirement is unchanged and still gets the whole registry, so this only affects sea ice.

One upgrade note. An existing installation already has the files on disk but no catalog entries for them, and a requirement that matches nothing produces no executions. So sea ice quietly stops solving until `ref providers setup --provider esmvaltool` is run again. The changelog fragment says so.

## Summary by CodeRabbit

* **New Features**
  * Reference data is now downloaded and catalogued together through the provider setup workflow.
  * Sea-ice diagnostics now automatically select the relevant OSI-450 Northern and Southern Hemisphere observations for comparisons.
  * Existing installations can populate missing reference catalog entries by rerunning provider setup.

* **Bug Fixes**
  * Improved handling when reference data or optional ingestion support is unavailable, with clear warnings instead of failures.


lewisjared:
@coderabbitai review

lewisjared:
@coderabbitai full review

lewisjared:
@coderabbitai review

lewisjared:
Fixed in 28571159. The warning now reads "No ESMValTool reference data has been fetched to <path>, so there is nothing to ingest" and points at `ref datasets fetch-data --registry esmvaltool-datasets`, which matches the remediation the diagnostic runner already suggests for the same missing tree. Added an assertion so the remediation cannot drift back to naming the command the hook runs under.

lewisjared:
@coderabbitai resolve

PR #861, opened 2026-08-12
fix(ilamb): upgrade ilamb3 and repair amoc-rapid

lewisjared:
There were some ilamb diagnostics that failed under a range of different conditions.
`mrsos-wangmao`, `snc-esacci` and `amoc-rapid`.

`mrsos-wangmao` and `snc-esacci` shared a root cause upstream. The existing ilamb3 derived time measures with `astype(float) * 1e-9`, which is only correct for `timedelta64[ns]`. cftime casts to ms instead of ns.

```
ilamb3 2026.2.19: mean time frequency = 0.0304 days
ilamb3 2026.6.15: mean time frequency = 30.4167 days
```

ilamb3 now uses `.dt.total_seconds()`, which is calendar-agnostic.

The upgrade retires two pins that existed only to work around ilamb3 issues it resolves: `xarray<2025.11` (its `integrate_space` units handling) and `scipy<1.16` ([statsmodels#9584](https://github.com/statsmodels/statsmodels/issues/9584), now satisfied by `statsmodels>=0.14.6`).

### amoc-rapid

Independently broken by the move onto obs4REF, in three layers:

1. ILAMB's `msftmz_to_rapid` selects the Atlantic with a hardcoded `isel(basin=0)`. The `RAPID-2023-1a` file this provider was moved onto ships the array already reduced to the RAPID section (`time`/`depth`/`lat`) and has no `basin` dimension, so it raised `ValueError: Dimensions {'basin'} do not exist`. The older `RAPID` file did have `basin`. 2. The analysis runs on the derived `amoc`, not the source `msftmz`, so `analysis_variable: amoc` was needed.
3. ilamb3 looks a single-source reference up by the *analysis* variable (`ref[variable_id]`), so the source key has to be the derived name. REF's rewrite now preserves the key the configure file chose instead of overwriting it with the on-disk variable.

## Checklist

- [x] Tests added

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved completion of three ILAMB models across the CMIP6 archive.
  - Fixed AMOC-RAPID processing for reference data without a basin dimension.
  - Preserved configured source mappings during observation data processing.
- **Compatibility**
  - Updated ILAMB3 support and relaxed outdated SciPy and xarray version restrictions.
- **Tests**
  - Added coverage for AMOC-RAPID transformations, including basin handling and expected latitude and depth selection.


lewisjared:
No reminting of the diagnostics were required

PR #866, opened 2026-08-13
fix(resources): attribute peak memory to the execution rather than the container

lewisjared:
## Description

Three linked fixes to per-execution resource measurement. The first is the load-bearing one.

**`exclusive` was a lie.** It was derived only from blocks measured inside the same process, via the process-global `_in_flight` registry. `LocalExecutor` spawns a fresh worker per task (`max_tasks_per_child=1`), so that registry is always empty in the worker and every row recorded `exclusive=True` — no matter how many sibling processes were saturating the same cgroup. `ResourcesReader` defaults to `exclusive_only=True` and therefore believed it had filtered contaminated rows when it had not.

The executor is the only party that knows its own concurrency, so it now says. `measure_resources` gained `cgroup_exclusive`, threaded through `execute_locally(exclusive=...)`, and the recorded flag is *declaration AND no in-process overlap* — the registry is kept as the second half, since it catches nested blocks a declaration cannot.

| Executor | Declares exclusive when |
| --- | --- |
| `LocalExecutor` | the pool has a single worker; an unrecognised pool is treated as concurrent |
| `HPCExecutor` | `max_workers_per_node == 1` |
| `SynchronousExecutor` | always — one execution at a time in one process |
| Celery | never; worker concurrency is set on the worker, not the submitter |

**Both peaks are now persisted.** `_PeakSampler` already tracked `cgroup_peak` and `proc_tree_peak` on every sample and threw one away. `ResourceUsage` now carries both, and they reach `resource_context` alongside `cgroup_exclusive_declared` — no migration needed. Which one to trust becomes a read-time decision, and a large divergence between them is itself the evidence that the cgroup was shared.

**`proc_tree` is now the default reported source.** Precedence is: cgroup if declared exclusive → proc_tree → cgroup (shared, and marked non-exclusive) → rusage. The principle is that proc_tree is correct by construction across both platforms and all three executors, while cgroup is correct only under a deployment assumption the library cannot verify. A shared cgroup reading still beats `getrusage`, which cannot be attributed to a block at all.

One consequence had to be followed through: `exclusive_only=True` previously dropped *every* non-exclusive row, which would now empty the profile of every diagnostic run by a pool. Exclusivity only disqualifies a *cgroup* reading, so the read-side filter is narrowed to that (`_is_attributable`). CLI warning text and `--include-shared` help updated to match.

### Notes for reviewers

- `test_workers_running_at_the_same_time_are_not_exclusive` was `xfail(strict=True)`, documenting exactly this bug; it now passes and the marker is gone. It was also passing for the wrong reason — its spawn children died on `ModuleNotFoundError` rather than on the exclusivity claim — so it and the new neighbour-contamination test use the file's existing `probe()` subprocess pattern instead.
- Running the whole `packages/climate-ref/tests/unit` directory in one process aborts in pyarrow at `test_solve_helpers.py::test_round_trip` on my 1.5 GB dev box. That reproduces on a stashed tree, so it is unrelated to this change; that file was run separately and passes.
- Not tested: the HPC and Celery paths end to end against a real scheduler or broker.

## Checklist

- [x] Tests added


lewisjared:
Superseded by #868 — GitHub closed this when the head branch was renamed from `t3code/14473093` to `fix/resource-exclusivity`. Same commits, same content.

PR #879, opened 2026-08-14
Close database connections on every entrypoint

lewisjared:
Audits how every entrypoint manages its database session and connection, then fixes the four problems that turned up. The CLI was already correct, and `Database.session_scope` holds up under concurrent threads, so neither changes here.

Celery workers opened a `Database` per task and never closed it. On SQLite the garbage collector eventually mopped up, so open files plateaued around ten. On Postgres it means a fresh connection for every result a worker handles. Workers now open one database per process and reuse it across tasks, keyed on the pid so a forked child opens its own rather than using connections it inherited. A `worker_process_shutdown` hook closes it. Measured over 100 tasks, open files drop from 10 to 1.

The three executors open a database of their own when the caller does not supply one, and never closed it. That path is not used in production, where the CLI always passes its database in, but it leaks for anyone constructing an executor by hand. Each now closes what it opened. The standalone alembic environment gets the same treatment.

`LocalExecutor.join` promised that outstanding executions are always marked failed-retryable before it returns or raises. Only the timeout path did this, so an ingestion failure mid-loop left the remaining rows with `successful=None` until the stale-execution reaper caught them six hours later. The promise now holds for every failure, and the pool is shut down on the way out.

Verified with three live experiments rather than by reading alone: file-descriptor counts across 100 simulated tasks, eight threads hammering `session_scope` on SQLite (0 errors), and a spy on `Database.close` confirming the CLI closes on success, on `typer.Exit`, and on an unexpected exception.

Adds three tests. `packages/climate-ref` and `packages/climate-ref-celery` pass in full at 1735 tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Failed in-flight executions are now recorded as failed and retryable when result processing encounters an error, allowing subsequent retries.
  * Improved cleanup of outstanding executions when local processing fails.
  * Database connections created automatically by executors and migrations are now closed reliably.

* **Performance**
  * Worker processes now reuse database connections where appropriate, reducing repeated connection setup.

* **Documentation**
  * Updated the changelog with details of retry handling and database connection reuse.

### DIFF
--- a/packages/climate-ref/src/climate_ref/executor/local.py
+++ b/packages/climate-ref/src/climate_ref/executor/local.py
@@ -1,6 +1,7 @@
 import concurrent.futures
 import multiprocessing
 import time
+import weakref
 from concurrent.futures import ProcessPoolExecutor
 from typing import Any
 
@@ -102,6 +103,8 @@ class LocalExecutor:
             config = Config.default()
         if database is None:
             database = Database.from_config(config, run_migrations=False)
+            # A database we opened is ours to close, so it does not outlive this executor.
+            weakref.finalize(self, database.close)
         self.n = n
 
         self.database = database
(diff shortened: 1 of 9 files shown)


packages/climate-ref/src/climate_ref/executor/local.py
```diff
@@ -287,6 +290,11 @@ def join(self, timeout: float) -> None:
 
                 # Wait for a short time before checking for completed executions
                 time.sleep(refresh_time)
+        except BaseException:
+            # Handle Ctrl-C and SystemExit exceptions to mark in-flight executions as retryable
+            self._fail_outstanding(results, t)
```
lewisjared, line 295: @fuchsi-huber This should resolve the Ctrl-C interrupt leaving the in flight diagnostics requirinng manual intervention

fuchsi-huber: Oh, nice! That is definitely a QoL improvement :)

PR #885, opened 2026-08-24
Update ESMValTool to v2.15 and reduce the size of the ESMValTool conda environment

bouweandela:
## Description

Use the ``esmvaltool-base`` package for a smaller conda-environment (about 2.3 GB).

See https://github.com/ESMValGroup/ESMValTool/pull/4595 for documentation of the package.

## Checklist

Please confirm that this pull request has done the following:

- [ ] Tests added
- [ ] Documentation added (where applicable)
- [ ] Changelog item added to `changelog/`

PR #886, opened 2026-08-28
Update the ozone diagnostics

lewisjared:
Updates the ozone diagnostics for the unified `ref/recipe_ref_ozone.yml` recipe that replaced `recipe_ref_ozone_cmip7.yml` in ESMValTool v2.15. Targets #885 because it depends on the recipe and lock file updates there.

- The base recipe's time ranges now drive the diagnostics. A shared `_prepare_recipe` helper replaces the per-diagnostic overrides.
- CMIP7 executions evaluate the historical experiment through 2021. The fabricated CMIP7 test data is extended to match, which moves the polar cap time series start to 1960 there.
- The obs4MIPs requirement demands coverage through 2021 for CMIP7 only. CMIP6 only needs it til 2015. I'm not sure if we need to reduce the coverage, but oh well.

Fixes the stale `recipes.txt` hashes for the renamed enso recipes and the sea-ice recipe.
The ENSO recipies will be fixed seperately.  

I need to rerun the regression baselines


lewisjared:
I've had to roll the diagnostics start date forward to 1960 due to a bug in the CMIP7 dataset extension.

lewisjared:
Closing in favour of a fresh PR. The branch history has been rebased onto use-esmvaltool-base and the old one carried duplicate commits.

PR #910, opened 2026-09-04
feat: render baseline diffs as a local HTML report

lewisjared:
Adds `ref test-cases diff --html-dir <dir>`, which compares every committed `manifest.json` against a base ref and writes a static HTML site to a local directory. 

Images render old and new side by side with a button that overlays them, text outputs get a coloured unified diff, and NetCDF and other binaries get a size delta and a link to each blob.

The rendering splits into `collect`, `analyse` and `render`. The rendering is performed with jinja. This adds it as a dependency.
An index file is also generated.

Verified against `fix/ozone-cmip7-recipe`, which renders nine case pages with images loading from the live store.


lewisjared:
@coderabbitai review

PR #912, opened 2026-09-04
feat: publish the baseline diff report and write the PR comment

lewisjared:
Adds `--upload <prefix>` and `--comment-output <path>` to `ref test-cases diff`, so the local report from #910 can be hosted and linked from a pull request comment. The comment becomes one table row per changed case, which is what replaces the capped markdown dump `scripts/ci/mint_diff.py` posts today.

Worth a close look:

- `ReportStore` is a new sibling of `NativeStore` rather than a method on it, because reports are keyed by path and `NativeStore` must stay content-addressed. Nothing in `store.py` changed.
- Key validation in `ReportStore.put` is the security-relevant bit. It layers an object-key character class over `safe_path`, the containment primitive `manifest.py` and `capture.py` already use.
- Every remote upload sets an explicit content type. Without it R2 serves the page as a binary download.
- Reports go to their own bucket, `ref-baselines-reports`, because an R2 token cannot be scoped to a prefix. A token that can write reports must not be able to overwrite baseline blobs.
- `ReportStore.preflight` deliberately mirrors `NativeStore.preflight` rather than sharing code with it, since `store.py` is out of scope here. The copy has already picked up a `BotoCoreError` branch that `store.py` lacks, so it is worth deciding whether the two should converge later.
- `render_comment` takes a `previous` argument that nothing supplies yet. The CI slice fills it in.

The marker `<!-- climate-ref-baseline-diff -->` is the contract the CI slice will use to find and edit the sticky comment.

No workflow or docs changes here. The bucket, host, token and GitHub environment do not exist yet, so nothing in CI calls this. That is the next slice.

### DIFF
--- a/changelog/912.feature.md
+++ b/changelog/912.feature.md
@@ -1,8 +1,4 @@
 `ref test-cases diff --upload <prefix>` publishes the report to a public object store,
 and `--comment-output <path>` writes the pull request comment markdown that links into it.
-The comment is one table row per changed case, so its size no longer grows with the diff.
 
-Reports go to their own store, configured by `REF_REPORT_STORE_URL`, `REF_REPORT_STORE_BUCKET`
-and `REF_REPORT_STORE_S3_ENDPOINT_URL`, with credentials read from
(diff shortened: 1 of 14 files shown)


lewisjared:
@coderabbitai review

PR #914, opened 2026-09-04
feat(esmvaltool): label metric values with the experiment

lewisjared:
Adds `experiment_id` to the `group_by` of the ESMValTool diagnostics that pin a single historical experiment, so their metric values carry the experiment as a dimension. ECS, TCR, TCRE, ZEC and the warming-level diagnostics already group on it.

The execution key changes for the affected diagnostics, so each has its `Diagnostic.version` bumped. Execution groups are keyed on the version as well as the key, so without the bump the old and new keys would coexist at the same version. The manifests advance `diagnostic_version` by hand rather than via `mint`, because the outputs did not change and the native baselines replay cleanly against them.

The regression baselines and test-data manifests are regenerated to match, which is nearly all of the diff.

## Summary by CodeRabbit

* **Improvements**
  * Diagnostic results now include the experiment as a dimension, making historical and `esm-hist` results distinguishable.
  * Updated diagnostics expose experiment and variant metadata consistently across climate sensitivity, ENSO, ozone, regional historical, sea-ice, fire-driver and cloud-radiative-effect analyses.
  * Refreshed execution grouping and diagnostic versions ensure new results are generated under the updated metadata keys.

* **Tests**
  * Updated catalogs, manifests, regression outputs and solver snapshots to reflect historical-experiment metadata.


lewisjared:
@coderabbitai review

lewisjared:
@coderabbitai review

PR #923, opened 2026-09-04
Mark tiny baseline diffs as noise rather than as changes

lewisjared:
Marks NetCDF statistics that only moved within floating point noise as a warning in the baseline diff report, rather than as a difference. A run reproduced on different hardware moves the last bits of nearly every cell, so almost every row was shaded red and the rows that actually moved were buried.

- A row whose largest relative difference is at or below `1e-9` is shaded amber and tagged `noise`, and its individual statistics are no longer emphasised.
- A move that cannot be measured, such as a changed shape or a changed NaN pattern, still reads as a real change.
- `Pair` gained a tolerance so a mean like `-6.98e-18 -> 8.37e-18` stops reading as moved. The tolerance is scaled by the base ref's own magnitude, so it means nothing for a variable that is entirely zero.
- The row class renamed from `moved` to `changed`, matching the three severities.

Two limits worth naming. The threshold is a module constant with no override, and I'm open to a different value, `1e-9` just leaves several digits of headroom under double precision. The file and case tallies are untouched, so a NetCDF whose every row is noise still counts as changed in the summary. That rollup felt like a separate change.

## Summary by CodeRabbit

- **New Features**
  - Baseline diff reports now identify numerical differences within the noise tolerance as **noise**.
  - Noise-only differences are shaded amber and labelled with a **noise** tag, helping distinguish them from material changes.
  - Differences exceeding the tolerance continue to be reported and highlighted as changed.

- **Bug Fixes**
  - Insignificant floating-point differences no longer appear as meaningful changes in NetCDF comparison reports.

### DIFF
--- a/changelog/923.improvement.md
+++ b/changelog/923.improvement.md
@@ -1,1 +1,5 @@
-Baseline diff reports now shade a NetCDF variable amber and tag it `noise` when every difference is within `1e-9` of the variable's own magnitude, instead of shading it as a real change. This keeps the last-bit wobble of a run reproduced on different hardware from burying the rows that actually moved.
+Baseline diff reports now shade a NetCDF variable amber and tag it `noise`
+when every difference is within `1e-9` of the base ref's own magnitude,
+instead of shading it as a real change.
+This keeps the last-bit wobble of a run reproduced on different hardware
+from burying the rows that actually moved.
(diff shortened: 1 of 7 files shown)


lewisjared:
@coderabbitai review

PR #930, opened 2026-09-17
feat: fix the datasets used for unit tests to the ids that are in the respective catalogs already

### LINKED ISSUE #737: ref test-cases fetch re-resolves from the search spec, not the committed catalog
## Describe the bug
When a catalog already exists, `ref test-cases fetch` reruns the search rather than just fetching the datasets in the catalog.
This has implications for reproducibility.

## Expected behavior
If a catalog exists everyone uses the same datasets, unless a user runs with `--regen`.

The fetch process should use the instance ids from the catalog instead of the general search query.

fuchsi-huber:
## Description

If possible, resolve datasets during `ref test-cases fetch` by the ids stored in the shipped catalogs instead of by their facets alone. To achieve this, the requests defined in the tests are pinned to the ids from the current catalog before fetching. This should ensure, that the used baselines don't shift unexpectedly due to new data being published to ESGF.

This closes https://github.com/Climate-REF/climate-ref/issues/737 . 

NOTE: 
* To enable consistent requests to the ESGF based on `instance_id`s, the obs4MIPs ids from pooch ingestion had to be changed to be consistent with the way ESGF creates them. 
* CMIP6 `instance_id`s had to be fixed to only take their first `activity_id` entry instead of all (`C4MIP CDRMIP` -> `C4MIP`).
* As pooch does not have unique ids, the corresponding requests pin facets (in particular the version) instead.
* CMIP7 requests translating data from CMIP6 are not pinned. The extra effort it would take did not seem reasonable considering we will remove this translational layer anyway.

Along the way:
* fixed fetch dry_run to actually use the same logic and outputs as the regular run. Made the output for skipping an existing catalog a debug message so a dry-run actually shows what would be done in a way a human can parse it.
* As I was touching the IDs anyways, I also unified the parsing of pmp climatology keys to match how the obs4mips parses the slot as `institution_short` rather than `institution_id` (which becomes `variant_label` in the catalog - maybe we should already call them that from the beginning?)
* Fixed CMIP7 conversion of datasets extended past the end of their CMIP6 source.

## Checklist

- [x] Tests added

## Summary by CodeRabbit

- **Breaking Changes**
  - Dataset identifiers now align with ESGF naming: redundant collection prefixes and nominal-resolution segments are removed.
  - CMIP6 identifiers use only the primary activity, while full activity metadata remains available.

- **New Features**
  - Existing catalogues are reused when fetching test data, with `--regen` available to refresh selections.
  - Catalogues now update when dataset metadata changes.

- **Bug Fixes**
  - Improved handling of chunked CMIP7 conversions with time bounds.
  - Missing datasets are reported clearly during catalogue-based fetching.

- **Documentation**
  - Added precise dataset identifier construction guidance.


packages/climate-ref-core/src/climate_ref_core/esgf/cmip7.py
```diff
@@ -348,6 +368,31 @@ def __init__(
...
+        Return this request unchanged: CMIP7 requests are not pinned.
+
+        A recorded CMIP7 dataset names nothing that identifies what it was built from. It
+        is converted locally, so ESGF has no id for it, and the CMIP6 dataset underneath is
+        no longer recognisable in the result: the version is the conversion's own,
+        ``table_id`` has no CMIP7 counterpart, and the Data Request renames variables
+        (``Lmon.mrsos`` and ``Emon.mrsol`` both become ``mrsol``). Recording the source id
+        alongside would fix that, but the whole CMIP6-to-CMIP7 conversion only stands in
+        until CMIP7 data is published, so these requests re-resolve their declared facets
+        and the pinning stays where the data is real.
```
lewisjared:
```suggestion
       CMIP7 data cannot be fetched from ESGF at the moment.
```

Rest of the comment isn't really useful.

packages/climate-ref-core/src/climate_ref_core/esgf/registry.py
```diff
@@ -101,7 +107,12 @@ def _parse_pmp_climatology_key(key: str) -> dict[str, Any]:
     PMP_obs4MIPsClims/{variable_id}/{grid_label}/{version}/{filename}
 
     Where filename is:
-    {variable_id}_mon_{source_id}_{institution_id}_{grid_label}_{time_range}_AC_{version}_{resolution}.nc
+    {variable_id}_mon_{source_id}_{inst_short}_{grid_label}_{time_range}_AC_{version}_{resolution}.nc
```
lewisjared: Rename to use `institution_short`. Trim the last sentence of the comment too

packages/climate-ref/src/climate_ref/datasets/obs4mips.py
```diff
@@ -217,10 +219,10 @@ def find_local_datasets(self, file_or_directory: Path) -> pd.DataFrame:
             self.version_metadata,
         ]
 
-        def _transform(item: str, value: Any) -> str:
-            return str(value).replace(" ", "") if item == "nominal_resolution" else str(value)
-
-        datasets = build_instance_id(datasets, drs_items, prefix=self.activity_id, transform=_transform)
+        # No prefix: ``activity_id`` leads the id and has just been set to this adapter's
+        # collection, so an obs4MIPs id is the one ESGF publishes the dataset under while
+        # an obs4REF id still says obs4REF.
```
lewisjared: The activity prefix was left over from when we had obs4REF and obs4MIPs datasets landing in the same database table. 

```suggestion
```

packages/climate-ref/src/climate_ref/datasets/obs4mips.py
```diff
@@ -164,14 +164,16 @@ class Obs4MIPsDatasetAdapter(DatasetAdapter):
 
     file_specific_metadata = ("start_time", "end_time", "path")
     version_metadata = "version"
-    # See ODS2.5 at https://doi.org/10.5281/zenodo.11500474 under "Directory structure template"
+    # The facets ESGF publishes obs4MIPs datasets under, so that a dataset's ``instance_id``
```
lewisjared, line 167: Ahh ok, so the directory layout <> instance_id for obs4MIPs. Annoying...

lewisjared (commented):
Nice. I believe this can be made non-breaking with a simple db migration to rewrite the instance_ids. I'll make a follow up PR to show how that could look.

I'll write you some instructions for how I rerun all the baselines. The CI action only supports running a single baseline at once

fuchsi-huber:
I have added another change to this PR, as I was touching all catalogs anyway:
* the storage_ids of the converted CMIP7 data is now reproducible

I have updated the manifests via `ref test-cases replay` and `ref test-cases run --from-slot --force-regen`. We should still run a recalculation of the baseline some time soon as some test-cases are missing baselines.

